PR #26, opened 2024-04-11
Bugfix for state increment array referencing in update_type=13

amfox37:
A bugfix for state increment array referencing in Update Type 13 for when we have SMAP observations.

Tested to give identical results as Update Type 10 over 8 cycles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -----------------------------
 
+## [v1.0.2] - 2024-04-11
+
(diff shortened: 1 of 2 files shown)


gmao-rreichle:
Thanks, @amfox37, for finding this bug and putting the PR together, and sorry for introducing the bug in my cleanup.  
I changed the label to 0-diff because there's no change in the results from the standard suite of nightly tests, which do not (yet) engage update_type=13.  It's also not really a science change in the new functionality (update_type=13).  Just a bug fix.

gmao-rreichle (approved):

weiyuan-jiang (approved):

PR #30, opened 2024-04-11
Update .gitignore to remove GEOSgcm_GridComp references

mathomp4:
Remove `GEOSgcm_GridComp` references in `.gitignore`.
(Minimal cleanup from #27.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,3 @@
 *#
 .#*
 **/CVS/
-
-/@GEOSgcm_GridComp
-/GEOSgcm_GridComp
-/GEOSgcm_GridComp@


gmao-rreichle (approved):

mathomp4:
Ahh. That was a subtle thing. The CI failed because it found an ancient branch called `mathomp4-patch-1` on GEOSgcm_GridComp. So old, the interfaces were not compatible. I've removed that branch from GEOSgcm_GridComp, so now all should work.

gmao-rreichle:
> Ahh. That was a subtle thing. The CI failed because it found an ancient branch called `mathomp4-patch-1` on GEOSgcm_GridComp. So old, the interfaces were not compatible. I've removed that branch from GEOSgcm_GridComp, so now all should work.

Thanks!  But why would a GEOSgcm_GridComp branch other than develop have anything to do with this PR?  Shouldn't CI use the branch that's listed in components.yaml of GEOSldas ?   I'm probably missing something, and if it works I'm not going to question it further.

mathomp4:
> > Ahh. That was a subtle thing. The CI failed because it found an ancient branch called `mathomp4-patch-1` on GEOSgcm_GridComp. So old, the interfaces were not compatible. I've removed that branch from GEOSgcm_GridComp, so now all should work.
> 
> Thanks! But why would a GEOSgcm_GridComp branch other than develop have anything to do with this PR? Shouldn't CI use the branch that's listed in components.yaml of GEOSldas ? I'm probably missing something, and if it works I'm not going to question it further.

This is my attempt to support multi-repo PRs. What the CI does is after cloning look for branches in other repos with the same name. If a user does that and they are dependent, then the CI will build instead of fail.

PR #40, opened 2024-05-02
remove restart options F and G

weiyuan-jiang:
These options are not necessary anymore. They were not well maintained and could create problems in the future.


gmao-rreichle:
Thanks, @weiyuan-jiang, for putting this together.  Have you or @biljanaorescanin tested the branch?  Looks good to me but to be safe we should test.

biljanaorescanin:
All nightly tests passed.

gmao-rreichle (approved):

PR #41, opened 2024-05-03
more optimal distribution of cubed-sphere grid

weiyuan-jiang:
When tiles are concentrated in one face, almost all processors are allocated to them , which may lead to too many processors on one face ( a processor may have 0 or 1 cell, which would crash regrider) . This PR tries to avoid that and stop when there are too many processors.

### DIFF
--- a/GEOSldas_App/preprocess_ldas_routines.F90
+++ b/GEOSldas_App/preprocess_ldas_routines.F90
@@ -2061,7 +2061,7 @@ contains
     integer :: n,typ,tmpint
     real ::  tmpreal
     integer :: avg_land,n0,local
-    integer :: i,s,e,j,k,n1,n2
+    integer :: i,s,e,j,k,n1,n2, s1, s2
     logical :: file_exist
     character(len=256):: tmpLine
     character(len=128):: gridname
@@ -2156,6 +2156,7 @@ contains
           face_land(k) = sum(landPosition(n1:n2)) 
           face(k) = nint(1.0*face_land(k)/total_land * N_proc)
           if ( face(k) == 0) face(k) = 1
+          if (face(k) > IMGLOB/2) face(k) = IMGLOB/2 ! make sure each process has at least 2 cells
        enddo
        
        ! now make sure sum(face) == N_proc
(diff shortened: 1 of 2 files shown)


GEOSldas_App/preprocess_ldas_routines.F90
```diff
@@ -2155,7 +2164,10 @@ subroutine optimize_latlon(fname_tilefile, N_proc_string, optimized_file, run_di
           n2 = k*IMGLOB
           face_land(k) = sum(landPosition(n1:n2)) 
           face(k) = nint(1.0*face_land(k)/total_land * N_proc)
```
gmao-rreichle, line 2166: @weiyuan-jiang: Is this safe without parenthesis or the use of `real()`?  The variables `face_land`, `total_land`, and `N_proc` are all integers. Does the multiplication with 1.0 always force all multiplications and divisions to be real?  Or is there any possibility that an integer division kicks in before the conversion to real?

GEOSldas_App/preprocess_ldas_routines.F90
```diff
@@ -2275,8 +2259,11 @@ subroutine optimize_latlon(fname_tilefile, N_proc_string, optimized_file, run_di
           endif
        enddo
        
-       if( k /=6 ) stop ("one or more processes may accross the face")
-      
+       if( k /=6 ) then
+          err_msg = 'one or more processes may across the face'
```
gmao-rreichle, line 2263: @weiyuan-jiang : I don't understand the error message.  Please clarify

weiyuan-jiang: > @weiyuan-jiang : I don't understand the error message. Please clarify
 If that condition (k/=6) is true, that means a processor ( more processors)  may have grid cells from different faces, which is not allowed.

GEOSldas_App/preprocess_ldas_routines.F90
weiyuan-jiang: > @weiyuan-jiang: Is this safe without parenthesis or the use of `real()`? The variables `face_land`, `total_land`, and `N_proc` are all integers. Does the multiplication with 1.0 always force all multiplications and divisions to be real? Or is there any possibility that an integer division kicks in before the conversion to real?

That 1.0 promotes all the integer type to real

gmao-rreichle (approved):

gmao-rreichle (changes requested):
@weiyuan-jiang: After I just approved the PR I noticed one more thing: It looks like `elemental function rms_cs()` isn't used.  If so, we should delete it.  Can you please take a quick look.

gmao-rreichle (approved):

Issue #48, opened 2024-05-23
Error when writing HISTORY output of ASNOW from ENSAVG Gridcomp


gmao-rreichle:
@saraqzhang reported an error when HISTORY output of ASNOW is requested from the [ENSAVG Gridcomp](https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/develop/GEOSens_GridComp/GEOS_EnsGridComp.F90).  This **only** works when the other fractional area coverage variables (FRSAT, FRUST, and FRWLT) are also (simultaneously) written out by HISTORY. 

If ASNOW output is requested **without** simultaneously writing FRSAT, FRUST, and FRWLT, I think the following calculation fails:
https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/9d6d6997de3eacda74e213d11f10f0eb00a5f149/GEOSens_GridComp/GEOS_EnsGridComp.F90#L3686

Note that ASNOW is treated differently from the other variables:
```
    call MAPL_GetPointer(import, FRSAT,  'FRSAT' ,rc=status)
    VERIFY_(status)
    call MAPL_GetPointer(import, FRUST,  'FRUST' ,rc=status)
    VERIFY_(status)
    call MAPL_GetPointer(import, FRWLT,  'FRWLT' ,rc=status)
    VERIFY_(status)
    ! for offline model , there is no 'ASNOW', recompute 
    ! call MAPL_GetPointer(import, ASNOW,  'ASNOW' ,rc=status)
    ! VERIFY_(status)

```
https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/9d6d6997de3eacda74e213d11f10f0eb00a5f149/GEOSens_GridComp/GEOS_EnsGridComp.F90#L2636C5-L2638C22

It's not clear to me exactly why ASNOW is treated differently.  The comment: `for offline model, there is no ASNOW` doesn't quite make sense to me.  ASNOW **is** an export in Catch_GridComp.

One solution may be to somehow "activate" FRSAT, FRUST, and FRWLT whenever ASNOW output is requested. 

Or perhaps we can treat ASNOW like the other variables? 

**If** the FR* variables are written simultaneously, we **should** compute ASNOW as follows to make sure that the areas sum up to 1:
[...]

PR #49, opened 2024-05-23
Support HISTORY output of ASNOW alone from ENSAVG Gridcomp

gmao-rreichle:
Addresses #48.

Solution: Calculate the ens average of ASNOW from scratch (like that of the other tile area fractions FRSAT, FRUST, and FRWLT). 

Note: Always allocating ens avg of fractional area variables in GEOS_EnsGridComp.F90 does not work without doing the same in GEOS_CatchGridComp.F90.

cc: @mathomp4, @weiyuan-jiang @saraqzhang

### DIFF
--- a/GEOSens_GridComp/GEOS_EnsGridComp.F90
+++ b/GEOSens_GridComp/GEOS_EnsGridComp.F90
@@ -2339,8 +2339,7 @@ contains
     real, dimension(:),pointer :: TPUNST,TPUNST_enavg
     real, dimension(:),pointer :: TPSAT,TPSAT_enavg
     real, dimension(:),pointer :: TPWLT,TPWLT_enavg
-    !real, dimension(:),pointer :: ASNOW,ASNOW_enavg
-    real, dimension(:),pointer :: ASNOW_enavg
+    real, dimension(:),pointer :: ASNOW,ASNOW_enavg
     real, dimension(:),pointer :: SHSNOW,SHSNOW_enavg
     real, dimension(:),pointer :: AVETSNOW,AVETSNOW_enavg
     real, dimension(:),pointer :: FRSAT,FRSAT_enavg
@@ -2633,9 +2632,8 @@ contains
     VERIFY_(status)
     call MAPL_GetPointer(import, FRWLT,  'FRWLT' ,rc=status)
     VERIFY_(status)
-    ! for offline model , there is no 'ASNOW', recompute 
-    ! call MAPL_GetPointer(import, ASNOW,  'ASNOW' ,rc=status)
-    ! VERIFY_(status)
+    call MAPL_GetPointer(import, ASNOW,  'ASNOW' ,rc=status)
+    VERIFY_(status)
     call MAPL_GetPointer(import, SNOWMASS,  'SNOWMASS' ,rc=status)
     VERIFY_(status)
     call MAPL_GetPointer(import, SNOWDP,  'SNOWDP' ,rc=status)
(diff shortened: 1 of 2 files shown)


GEOSens_GridComp/GEOS_EnsGridComp.F90
```diff
@@ -2924,11 +2924,11 @@ subroutine Collect_land_ens(gc, import, export, clock, rc)
     VERIFY_(status)
     call MAPL_GetPointer(export, AVETSNOW_enavg,  'AVETSNOW' ,rc=status)
     VERIFY_(status)
-    call MAPL_GetPointer(export, FRSAT_enavg,  'FRSAT' ,rc=status)
+    call MAPL_GetPointer(export, FRSAT_enavg,  'FRSAT' , alloc=.true., rc=status)   ! always allocate to support ASNOW_enavg calc below
```
gmao-rreichle, line 2927: @mathomp4, @weiyuan-jiang: 
Would it make sense to determine whether we really need to allocate  the FR* variables?  If the following works, it would save a small amount of memory at the expense of adding an if statement and a few extra lines of code.

```
if (associated(ASNOW_enavg)) then

  call MAPL_GetPointer(export, FRSAT_enavg,  'FRSAT' , alloc=.true., rc=status)   ! always allocate to support ASNOW_enavg calc below
  VERIFY_(status)
  call MAPL_GetPointer(export, FRUST_enavg,  'FRUST' , alloc=.true., rc=status)   ! always allocate to support ASNOW_enavg calc below
  VERIFY_(status)
  call MAPL_GetPointer(export, FRWLT_enavg,  'FRWLT' , alloc=.true., rc=status)   ! always allocate to support ASNOW_enavg calc below
  VERIFY_(status)

else

  call MAPL_GetPointer(export, FRSAT_enavg, 'FRSAT' ,rc=status)
  VERIFY_(status)
  call MAPL_GetPointer(export, FRUST_enavg,  'FRUST' ,rc=status)
  VERIFY_(status)
  call MAPL_GetPointer(export, FRWLT_enavg,  'FRWLT' ,rc=status)
  VERIFY_(status)

end if
```
This may not work and there's probably a more elegant way to make the allocation conditional.  And perhaps it's silly to worry about always or only sometimes allocating these three exports.  Thoughts?

mathomp4: Well, one possibility (but would need testing) is first do the GetPointer for `ASNOW_enavg`. 

Then maybe this could work:

```

call MAPL_GetPointer(export, FRSAT_enavg,  'FRSAT' , alloc=associated(ASNOW_enavg), rc=status)

```

That might not though (maybe it's not kosher Fortran?). But you could have a logical and then:

```

logical :: ASNOW_enavg_associated
...

ASNOW_enavg_associated = associated(ASNOW_enavg)

call MAPL_GetPointer(export, FRSAT_enavg,  'FRSAT' , alloc=ASNOW_enavg_associated, rc=status)

```

Would want to test both just in case!

gmao-rreichle: Thanks, @mathomp4, for the hints.  Re. the suggestion:
```
ASNOW_enavg_associated = associated(ASNOW_enavg)
call MAPL_GetPointer(export, FRSAT_enavg,  'FRSAT' , alloc=ASNOW_enavg_associated, rc=status)
```
Would this still work if **only** FRSAT but **not** ASNOW is written?  Then we would have alloc=.false.  Not sure what MAPL_GetPointer() would do with the "alloc" spec when at the same time FRSAT output is requested in HISTORY.rc.

mathomp4: @gmao-rreichle Hmm. you might be right. Staring at [the code](https://github.com/GEOS-ESM/MAPL/blob/df6786ca32f75b4848743301f5a31e6a6288622e/base/GetPointer.H#L78-L85), it seems like if `alloc` is present, it wins. So passing in `alloc=.false.` could maybe short circuit things.

I'm going to mention @bena-nasa and @atrayano here. They might know better, but my guess is the easy way is to just always `alloc` the needed fields or do the big if-block like you have up there.

weiyuan-jiang: Actually, there are more dependence. ASNOW_enavg depends  on FRSAT_enavg, ..., and FRSAT_enavg depends on FRSAT in Catch_GridComp.  I think it is too tricky to enforce the dependence chain.  We can simply write out error message if ASNOW_enavg is requested but not those dependent variables.

gmao-rreichle: > Actually, there are more dependence. ASNOW_enavg depends on FRSAT_enavg, ..., and FRSAT_enavg depends on FRSAT in Catch_GridComp. I think it is too tricky to enforce the dependence chain. We can simply write out error message if ASNOW_enavg is requested but not those dependent variables.

Thanks, @weiyuan-jiang, for the input. Much appreciated.  I implemented a different solution, which I hope will work.  

@saraqzhang : When you get a chance, please try this branch to see whether you can now output just ASNOW (without also writing FRSAT, FRUST, and FRWLT.

saraqzhang: I tested the branch ( output ASNOW without FRSAT, FRUST, and FRWLT), it fixed the problem,  and the difference is on the order of 1e-7   if comparing ASNOW field from this output to ASNOW before ( dependent of  FRSAT, FRUST, and FRWLT).

gmao-rreichle:
successfully 0-diff tested for Intel (incl debug) by @gmao-rreichle

biljanaorescanin (approved):

PR #56, opened 2024-10-02
Fixes #55. ifdef out HDF4 readers

### LINKED ISSUE #55: Does GEOSldas still require HDF4?
A query for @gmao-rreichle. Recently, I've been doing some work to try and get `ifx` support into GEOS, Baselibs, and the GEOS CI since by the end of this year Intel will be killing `ifort` as a Fortran compiler.

The main issue at the moment for LDAS is that the FORTRAN 77 interfaces to HDF4 are deprecated[^1] and even if you *force* HDF4 to build them (as I do in Baselibs) even *that* does not work with `ifx`. So for now in my testing, if you build Baselibs with `ifx`, HDF4 is not built.

Now, I believe eventually the HDF Group plans to add "back" Fortran support with modern Fortran (see https://github.com/HDFGroup/hdf4/discussions/408 by @derobins), so might just be a "temporary" issue.

But perhaps the HDF code in the LDAS (namely it seems to be [an AMSR-E obs reader](https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/b359a1fc13fea6a23e4a9c0c09b6a61951a6159d/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90#L93)?) isn't *needed* but just kept for backwards-looking work? If so, then maybe if there is a point in time where we have to move to `ifx` but HDF4 4.4.0 is not out yet, we can, if nothing else, have CMake check for HDF4 support and if not found, `#ifdef` out the code?

[^1]: Note: They might even be *removed* in HDF4 4.3.0. I haven't actually tried updating and testing that yet. So this issue might be an "all-compiler" issue rather than just "ifx".

mathomp4:
### Note: The change seems more exciting than it is because my editor strips off trailing whitespace. To better see the change, use the "no whitespace" option: https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/56/files?w=1

---

Closes #55 

This PR adds `#if 0` blocks around code that uses the HDF4 Fortran interface. As stated in #55, HDF4 (at the moment) no longer supports a Fortran interface (due to implementation bugs in the library)

Per https://github.com/GEOS-ESM/GEOSldas_GridComp/issues/55#issuecomment-2284618747 by @gmao-rreichle, a solution is to just `#if 0` the offending code so it remains, but isn't compiled (thus removing the need for HDF4).

I *think* this is all that's needed, and it does build with `ifx` but I'm a bit scared as that was a lot of code I ifdef'd. I think in the end it turns off support for:

```fortran
    case ('ae_l2_sm_a', 'ae_l2_sm_d')
...
    case ('MOD10C1','MYD10C1')
```

and if those are still assimilated...

So, I'll keep it Draft until @biljanaorescanin can do a test.

### DIFF
--- a/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
+++ b/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
@@ -5125,6 +5125,11 @@ contains
 
   ! *****************************************************************
 
+! NOTE: This code uses HDF4. HDF4 currently does not support
+!       building Fortran interfaces due to bugs. For now we
+!       comment out the code that uses HDF4 but keep it in
+!       the code base for future reference.
+#if 0
   subroutine read_obs_MODIS_SCF(                                   &
        date_time, dtstep_assim, N_catd, tile_coord,                &
        tile_grid_d, N_tile_in_cell_ij, tile_num_in_cell_ij,        &
@@ -5576,6 +5581,7 @@ contains
 
 
   end subroutine read_obs_MODIS_SCF
+#endif
 
   ! *******************************************************************************************************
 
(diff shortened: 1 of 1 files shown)


mathomp4:
Throwing in a blocking label as we need one of the obs this turns off.

gmao-rreichle:
Obsolete.  Closing.  C-bridge for hdf4 implemented in #194

PR #69, opened 2024-12-03
Update ASCAT obs reader to work with original and revised file name template

amfox37:
The bufr ASCAT soil moisture product we assimilate is no longer available with the filename format we have historically downloaded/used, although the file content is unchanged.

Example name change:
M02-ASCA-ASCSMO02-NA-5.0-20160101005700.000000000Z-20160101023821-1344690.bfr ->
W_XX-EUMETSAT-Darmstadt,SOUNDING+SATELLITE,METOPA+ASCAT_C_EUMC_20160101005700_47740_eps_o_250_ssm_l2.bin

From 1 June 2023 the names used will switch from the old (ASCA-ASCSMO02)  to the new (W_XX-EUMETSAT-Darmstadt) format.

Code modifications required:

1. Conditionally parse timestamp from the filename in different places depending if the old or new file name format
2. Check the parsed timestamp string only contains numeric characters
3. Check the timestamp has valid year, month and day integers
4. Increase the length of fname_list to accommodate new (very!) long filename
5. Increase length of tempstr to allow reading this new long filename
6. Add explicit format specifier in read to deal with comma in filename
7. Remove obs_param_nml%name entry from default as this isn't used (obs filename is read from %flist_name)

Honestly, the checks might be overkill as things will crash very quickly if anything other than 14 numeric characters are read into date_time_tmp (and the chances of those 14 numeric characters being anything other than a legitimate timestamp are minimal).


gmao-rreichle:
All LDAS nightly tests passed.

gmao-rreichle (approved):

PR #73, opened 2024-12-18
Time ranges for observation availability

amfox37:
Have added hard coded dates to the read_obs subroutines read_obs_sm_ASCAT_EUMET(), read_obs_SMAP_halforbit_Tb(), read_obs_SMOS() and read_obs_MODIS_SCF() so they do not try to read observations (or observation list files) for dates before they are available. i.e. not before 2015-3-31 in the case of SMAP. 
There are start dates for SMAP Tb, SMOS, MODIS SCF and both start and end dates for ASCAT SM.
I'm not sure we need the warning message, which will clog up the log file?


GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -4497,6 +4522,17 @@ subroutine read_obs_SMOS( date_time, N_catd, this_obs_param,      &
     
     found_obs = .false.
 
+    ! determine operating time range of sensor
+
+    date_time_obs_beg = date_time_type(2010, 5,24, 0, 0, 0,-9999,-9999)
```
gmao-rreichle, line 4527: @gmao-qliu:  What is the earliest date for which we have good-quality SMOS obs?  My memory suggests that useful SMOS science data only start in July 2010.

gmao-qliu: We have continuous valid data starting from 2010/05/24, so the date seems correct. We used 2010/06/01 onward for estimating the obs scaling parameters for SMAP in the past.

gmao-rreichle:
successfully tested by @gmao-rreichle

PR #76, opened 2025-01-17
Add CYGNSS soil moisture observation reader

amfox37:
This PR aims to add ability to read CYGNSS L3 soil moisture daily and subdaily soil moisture v3.2 observations from daily files available from: https://podaac.jpl.nasa.gov/dataset/CYGNSS_L3_SOIL_MOISTURE_V3.2
Subdaily observations represent 6 hour long time slices and are assimilated at 3z, 9z, 15z, 21z. Daily observations are assimilated at 12z. NB It is not anticipated both subdaily and daily observations will be jointly assimilated, although the code alllows this. 
There is no capability to assimilate two subdaily time slices simultaneously, so assimilation windows longer than 6hrs will error out.
Individual observation "sigma" values are currently read, but not used. errstd from nml is used. 
A fixed mask is read and applied  with obs removed when any flags are set.
This mask is on the same 36km grid as the 36km resolution observations. If the observations are not on this grid it will error out.
For now, employs scale_obs_sfmc_zscore(), which uses scaling parameters on lat/lon grid in nc4 file (as in ASCAT_MET* species).


GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -2145,7 +2145,511 @@ subroutine read_obs_sm_ASCAT_EUMET(                                         &
...
+       ierr = nf90_get_var(ncid, sm_subd_varid,    sm_subd)
+       ierr = nf90_get_var(ncid, sigma_subd_varid, sigma_subd)
+
+       ! find the time slice that corresponds to the assimilation window
+       idx = -1
+       do i = 1, N_timeslices
+           if ((date_time%hour ==  3 .and. timeintervals(1,i) == 0.0 ) .or.            &
+               (date_time%hour ==  9 .and. timeintervals(1,i) == 0.25) .or.            &
+               (date_time%hour == 15 .and. timeintervals(1,i) == 0.5 ) .or.            &
+               (date_time%hour == 21 .and. timeintervals(1,i) == 0.75)         ) then
```
gmao-rreichle, line 2415: This seems incompatible with assimilation windows that are centered on 3z, 6z, ... (as in [1:30,4:30], [4:30,7:30], ...)
Also, wouldn't it be easier to identify the idx value from H[xx], as in:  
```
if     H03 then
   idx=1
elseif H09 then
   idx=2
...
```

GEOSldas_App/LDASsa_DEFAULT_inputs_ensupd.nml
```diff
@@ -2363,9 +2363,87 @@ obs_param_nml(53)%xcorr          = 0.
...
+obs_param_nml(54)%scale          = .false.
+obs_param_nml(54)%getinnov       = .false.
+obs_param_nml(54)%RTM_ID         = 0
+obs_param_nml(54)%bias_Npar      = 0
+obs_param_nml(54)%bias_trel      = 864000
+obs_param_nml(54)%bias_tcut      = 432000
+obs_param_nml(54)%nodata         = -9999.
+obs_param_nml(54)%varname        = 'sfmc'
+obs_param_nml(54)%units          = 'm3 m-3'
+obs_param_nml(54)%path           = '/discover/nobackup/amfox/CYGNSS_obs/'
```
gmao-rreichle, line 2389: Let's see if we can put the data on a project disk, perhaps where we keep ASCAT

GEOSldas_App/LDASsa_DEFAULT_inputs_ensupd.nml
```diff
@@ -2363,9 +2363,87 @@ obs_param_nml(53)%xcorr          = 0.
...
+obs_param_nml(54)%name           = ''
+obs_param_nml(54)%maskpath       = ''
+obs_param_nml(54)%maskname       = ''
+obs_param_nml(54)%scalepath      = ''
+obs_param_nml(54)%scalename      = ''
+obs_param_nml(54)%flistpath      = ''
+obs_param_nml(54)%flistname      = ''
+obs_param_nml(54)%errstd         = 0.1
```
gmao-rreichle, line 2397: The 0.1 m3/m3 value for errstd seems rather large and would limit the impact of CYGNSS sm DA considerably.  Is this the recommended (default) value from CYGNSS?  For ASCAT, we use 0.04 m3/m3.  I recognize that this is just a default and that the obs reader pulls the errstd value from the CYGNSS obs file.  We might want to adjust the default here based on what's in the CYGNSS files and/or based on what we'll find to work once we have some CYGNSS DA results

GEOSldas_App/LDASsa_DEFAULT_inputs_ensupd.nml
```diff
@@ -2363,9 +2363,87 @@ obs_param_nml(53)%xcorr          = 0.
...
+obs_param_nml(54)%std_normal_max = 2.5
+obs_param_nml(54)%zeromean       = .true.
+obs_param_nml(54)%coarsen_pert   = .false.
```
gmao-rreichle, line 2400: I recognize that coarsen_pert is set to false for SMAP Tb obs in the DEFAULT nml file, but in L4_SM ops it's set to true.  Maybe specify true here?  (I'd need to re-familiarize myself with what exactly happens if this is set to true.)

GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -2145,7 +2145,511 @@ subroutine read_obs_sm_ASCAT_EUMET(                                         &
...
+    N_obs = 0
+
+    do i = 1, N_lon
+        do j = 1, N_lat
+            if (tmp_sm(i,j) .ne. this_obs_param%nodata .and.         &
+                small_SM_range_flag(i,j) .ne. 1        .and.         &
+                poor_SMAP_flag(i,j)      .ne. 1        .and.         &
+                high_ubrmsd_flag(i,j)    .ne. 1        .and.         &
+                few_obs_flag(i,j)        .ne. 1        .and.         &
+                low_signal_flag(i,j)     .ne. 1 ) then
```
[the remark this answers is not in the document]
gmao-rreichle: I suppose we could also just use something like:
good_flag_value = 255
if ...
   flag == good_flag_value
end

That would make it explicit that we have to have whatever the "good" flag value is (as opposed to avoiding a bad value, of which in future there may be more than one)

GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -2145,7 +2145,511 @@ subroutine read_obs_sm_ASCAT_EUMET(                                         &
...
+    HHcnt = 0
+
+    if (date_time_low%hour < 3 .and. date_time_up%hour >= 3) then
+       HH03 = .true.
+       HHcnt = HHcnt + 1
+    end if
+    if (date_time_low%hour > 21 .and. date_time_up%hour >= 3) then  ! wrap-around from previous day
+       HH03 = .true.
+       HHcnt = HHcnt + 1
+    end if
```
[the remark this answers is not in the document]
amfox37: OK - I'll give that a go. I'm also trying to figure out the single time slice read from the netCDF file (ie using idx iwith nf90_get_var).

GEOSldas_App/LDASsa_DEFAULT_inputs_ensupd.nml
```diff
@@ -2363,9 +2363,87 @@ obs_param_nml(53)%xcorr          = 0.
...
+obs_param_nml(54)%scale          = .false.
+obs_param_nml(54)%getinnov       = .false.
+obs_param_nml(54)%RTM_ID         = 0
+obs_param_nml(54)%bias_Npar      = 0
+obs_param_nml(54)%bias_trel      = 864000
+obs_param_nml(54)%bias_tcut      = 432000
+obs_param_nml(54)%nodata         = -9999.
+obs_param_nml(54)%varname        = 'sfmc'
+obs_param_nml(54)%units          = 'm3 m-3'
+obs_param_nml(54)%path           = '/discover/nobackup/projects/gmao/smap/SMAP_Nature/CYGNSS/'
```
amfox37, line 2389: CYGNSS observations moved to SMOS and ASCAT_EUMETSAT location

GEOSldas_App/LDASsa_DEFAULT_inputs_ensupd.nml
amfox37: Updated. See comment below

GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -2145,7 +2145,512 @@ subroutine read_obs_sm_ASCAT_EUMET(                                         &
...
+    ierr = nf90_get_var(ncid, lon_varid, longitudes)
+
+    idx = -1
+
+    ! read either subdaily or daily soil moisture and sigma variables
+    
+    if (dt_obs == 6) then
+       do i = 1, N_timeslices
+          if ((obs_hour ==  3 .and. (abs(timeintervals(1,i) - 0.0 ) < tolerance)) .or.            &
+              (obs_hour ==  9 .and. (abs(timeintervals(1,i) - 0.25) < tolerance)) .or.            &
```
gmao-rreichle, line 2402: Does timeintervals(1,:) refer to the **start** times of the observation interval?  
If so, I think we assimilate the obs from a "forward-looking" obs interval during a "backward-looking" assimilation window.  For example, if date_time=1:30z and the assimilation window is [0,3], then we set dt_obs=3 and assimilate obs taken during [3,9].  Can this be improved by changing the < and >= signs to <= and > in the calc of dt_obs?  I'm not really sure, but it's probably worth thinking about a bit more

amfox37: Yes, timeintervals(1,:) refers to the start times of the observation interval. 
I think the behaviour at the moment for 3hr assimilation windows centered on 1:30z and 4:30z is that it will assimilate all the obs from the first time slice (0 - 6hrs) in the first assimilation window centered on 1:30z and not in the second assimilation window centered on 4:30z. Changing < and >= signs to <= and > would reverse the behaviour.

gmao-rreichle:
After https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/76/commits/e2747183fc6ab31675c121e8bd2ca48c374c9a50

- CYGNSS reader functionality successfully tested by @amfox37 
- Passed LDAS nightly 0-diff tests (@gmao-rreichle)

gmao-rreichle (approved):

PR #80, opened 2025-01-28
run on cas when ntasks-per-node is less than 46

weiyuan-jiang:
We probably can wait until all cas machines are on SLES15. Then we can eliminate all "BUILT_ON_SLES15" logic.


mathomp4:
Also, the skylakes are now gone at NCCS, so if you have any code for that, it should be removed too.

gmao-rreichle:
PR is obsolete, now that Discover is fully on SLES15

PR #81, opened 2025-01-28
        updates for integrated setup of coupled land/atm DAS

saraqzhang:
Updates are implemented to integrate ldas_setup into the coupled  land/atm DAS setup procedure.  In the coupled land/atmosphere config, the ADAS setup script (fvsetup) calls ldas_setup with optional command line inputs, so that LDAS experiments in the coupled land/atmosphere DAS are set up according to ADAS experiment specifications, incl. exp path, exp id, begin date and gid.  

Changed nomenclature of met forcing files ("Nx+-" --> "bkg.lfo_*").

Successfully conducted LDAS regression tests (@gmao-rreichle, 14 April 2025)

To be done at a later time in a new PR:
- Clean up "Nx+-" nomenclature.
- Verify consistency of ldas_setup changes with overall design of ldas_setup. 
- Eliminate sample exeinp files for coupled land/atm DAS config  by hard-coding some rc variables in ldas_setup for this use case.  See https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/81/files#r2040075497.

Related PRs: 
https://github.com/GEOS-ESM/GEOSadas/pull/327
https://github.com/GEOS-ESM/GEOSgcm_App/pull/707


GEOSldas_App/ldas_setup
```diff
@@ -153,6 +154,9 @@ class LDASsetup:
 
         for key in rqdExeInpKeys :
             assert key in self.rqdExeInp,' "%s" is required in the input file %s' % (key,self.exeinpfile)
+        if cmdLineArgs['nymdb' ] != 'None' :
+            self.date  = f"{self.nymdb} 210000"
```
gmao-rreichle, line 158: I think we will need to pass in HHMMSS as well, or at least HH.  Should this ever run in FP, we cannot hardwire 21z.

GEOSldas_App/ldas_setup
```diff
@@ -168,8 +172,16 @@ class LDASsetup:
         self.ladas_coupling = int(self.rqdExeInp.get('LADAS_COUPLING',0))
         self.adas_expdir =''
         if self.ladas_coupling > 0:
-           assert 'ADAS_EXPDIR'  in self.rqdExeInp, " need ADAS_EXPDIR  in the input file %s" %(self.exeinpfile)
-           self.adas_expdir = self.rqdExeInp['ADAS_EXPDIR']
+           self.adas_expdir = os.path.dirname(self.exphome)
+           self.rqdExeInp[ 'ADAS_EXPDIR'] = self.adas_expdir
+           self.adas_expid = os.path.basename(self.adas_expdir)
+           self.rqdExeInp[ 'MET_TAG' ] = self.adas_expid + '__Nx+-'
```
gmao-rreichle, line 178: Do we really want to edit ldas_setup with '__Nx+-' knowing that this is already being changed?  Can we go straight to the new "bkg" location of the lfo files?

GEOSldas_App/ldas_setup
```diff
@@ -1112,6 +1124,14 @@ class LDASsetup:
...
                 for line in fileinput.input(tmprcfile,inplace=True):
                     print (line.rstrip().replace('GEOSldas_expid',self.rqdExeInp['EXP_ID']))
+                if (self.ladas_coupling == 1) :
+                    for line in fileinput.input(tmprcfile,inplace=True):
+                        print (line.rstrip().replace("#                'catch_progn_incr'","   'catch_progn_incr'"))
+                    for line in fileinput.input(tmprcfile,inplace=True):
+                        print (line.rstrip().replace('catch_progn_incr.ref_time:    000000,','catch_progn_incr.ref_time:    013000,'))
+                    for line in fileinput.input(tmprcfile,inplace=True):
+                        print (line.rstrip().replace("'SMAP_L4_SM_gph'","# 'SMAP_L4_SM_gph'"))
+
```
gmao-rreichle, line 1134: It might make more sense to add a new template of GEOSldas_HIST.rc that is specific to the LADAS, rather than edit GEOSldas_HIST.rc in this way

GEOSldas_App/ldas_setup
```diff
@@ -1710,6 +1728,11 @@ def parseCmdLine():
         help='replace computing/sponsor account in batinp file',
         type=str, default='None'
         )
+    p_setup.add_argument(
+        '--nymdb',
+        help='replace BEG_DATE in expinp file ',
```
gmao-rreichle, line 1733: Not sure I understand the comment.  Should this read:
'replace**s** BEG_DATE in expinp file'    ?? (without the s in "replaces", it reads like an instruction, when I think this help text describes the effect of specifying --nymdb
Also, it looks like there's a typo:  "exeinp" instead of "expinp"?

GEOSldas_App/ldas_setup
```diff
@@ -168,8 +172,16 @@ class LDASsetup:
...
-           self.adas_expdir = self.rqdExeInp['ADAS_EXPDIR']
+           self.adas_expdir = os.path.dirname(self.exphome)
+           self.rqdExeInp[ 'ADAS_EXPDIR'] = self.adas_expdir
+           self.adas_expid = os.path.basename(self.adas_expdir)
+           self.rqdExeInp[ 'MET_TAG' ] = self.adas_expid + '__Nx+-'
+           if self.ladas_coupling == 1:
+                 self.rqdExeInp[ 'EXP_ID' ] = self.adas_expid + '_LDAS'
+           if self.ladas_coupling == 2:
+                 self.rqdExeInp[ 'EXP_ID' ] = self.adas_expid + '_LDAS4ens'
+                 self.rqdExeInp[ 'MET_PATH' ] = self.adas_expdir +'/atmens/ensdiag/mem'
```
gmao-rreichle, line 183: See comment above -- avoid adding "./ensdiag/" when we are already changing to "bkg"

gmao-rreichle (commented):
@saraqzhang : I added a few inline comments that probably require further discussion at our next tag-up

GEOSldas_App/ldas_setup
saraqzhang: will add HHMMSS.

GEOSldas_App/ldas_setup
saraqzhang: I will leave  Nx+-  option in till we get bkg updates in  and tested

GEOSldas_App/ldas_setup
saraqzhang: can we add a new template  for central and  one for atmens  in  GEOSldas_App/ , so that they will be installed along with GEOSldas_HIST.rc  in  install-SLES15/etc/  by adas build.

GEOSldas_App/ldas_setup
saraqzhang: spot on.   maybe we can say "supersedes BEG_DATE in expinp file "

GEOSldas_App/ldas_setup
saraqzhang: I will leave this in till we get bkg updates in and tested

gmao-rreichle: I guess the question is whether this PR needs to be merged before we can switch to the "bkg" convention for the lfo inputs to the  LDAS.  My preference would be to avoid merging this PR before we make the switch, but I'm not sure if that's workable

GEOSldas_App/GEOSldas_HISTdet.rc
```diff
@@ -0,0 +1,70 @@
...
+#
+#  (1)  The "catch_progn_incr" is output is the ensemble average. 
+#  (2)  The "catch_progn_incr" output is in tile space.  Its definition is generic
+#
+##################################################################################
+
+VERSION: 1
+EXPID:   GEOSldas_expid
+
+COLLECTIONS:
```
gmao-rreichle, line 15: I think we need to add the usual diagnostic `tavg[*]_2d_lnd_Nx`, and `inst[*]_2d_lndfcstana_Nx` collections that we will want to output from the LDAS whenever it's running in the coupled land-atm DA system.  

The time step could be hourly (as in FP and reanalysis) or perhaps 3-hourly (as in SMAP L4).  

Note that the aforementioned collection names are from the MERRA-2 era and reflect what is in FP now.   For M21C, there's slightly different naming convention.  I assume the newer convention will be adopted in FP when it transitions to the forthcoming "v12" (181-level) GCM, so maybe it's best to use the newer collection naming convention here.

GEOSldas_App/GEOSldas_HISTens.rc
```diff
@@ -0,0 +1,1083 @@
...
+#
+#  - The "catch_progn_incr" output is in tile space, which must be the same for
+#       GEOSldas and ADASens.
+#
+#
+##################################################################################
+
+EXPID:  GEOSldas_expid
+
+COLLECTIONS:
```
gmao-rreichle, line 25: See comment above -- add "lnd" and "lndfcstana" collections (ens avg)

GEOSldas_App/ldas_setup
```diff
@@ -1079,7 +1092,7 @@ class LDASsetup:
 
         # DEFAULT rc files
         default_rc = glob.glob(etcdir+'/GEOSldas_*.rc')
-        assert len(default_rc)==4
+        assert len(default_rc)==6
```
gmao-rreichle, line 1095: Does this change work for a stand-alone (offline) config of LDAS?  It's unclear to me how we can just change from 4 to 6 without breaking something for the offline system.  I'm probably misunderstanding this

GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -5405,8 +5460,11 @@ subroutine parse_G5DAS_met_tag( met_path_in, met_tag_in, date_time, &
           
           use_Predictor = .true.
 
-       else
-          
+       elseif (tmp_tag(ii)(1:3)=='bkg') then
+
+          use_bkg = .true. 
+
```
gmao-rreichle, line 5466: I have to think more about this, but can't we simply replace "use_Predictor" with "use_bkg"?  The purpose is the same... use lfo output from the Predictor segment of the GCM.
There would probably be issues with backward compatibility if we eliminated "use_Predictor" and the `Nx+-` syntax, but we may not need backward compatibility in this case.  
This requires more thought

GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -6646,6 +6704,7 @@ program ut_parse_G5DAS_met_tag
   met_tag_in_vec = (/                              &
        'gcmexpname' ,                          &
        'gcmexpname__Nx+-' ,                    &
+       'gcmexpname__bkg' ,                     &
        'gcmexpname__precCORRPREC',             &
        'gcmexpname__precCORRPREC__Nx+-' ,            &
        'gcmexpname__Nx+-__precCORRPREC' ,            &
```
gmao-rreichle, line 6710: we  may need the option to run bkg **with** precip corr, as in:   `'gcmexpname__bkg__precCORRPREC'` and/or `'gcmexpname__precCORRPREC__Nx+-'`  (not sure whether both options for ordering the substrings are needed (or what the difference may be)
requires more thought

GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -3768,6 +3809,17 @@ subroutine get_GEOS( date_time, force_dtstep, met_path, met_tag,             &
...
+      if (use_bkg) then
+
+         do j=1,N_GEOSgcm_vars
+
+             GEOSgcm_defs(j,3) = G5LADAS_defs(j,3) 
+             GEOSgcm_defs(j,4) = G5LADAS_defs(j,4)
+
+         end do
+
+      end if
```
gmao-rreichle, line 3822: This should be handled in the if-block in lines 3783 etc

gmao-rreichle (commented):
@saraqzhang, I added a few inline comments/suggestions/questions.  Please take a look, thanks

GEOSldas_App/GEOSldas_HISTdet.rc
saraqzhang: those diag collections can be added to the template.  Because they are lat/lon outputs, more modification is needed in setup, we will need to insert the grid name ( from setup info) in the specification of "regrid_name", probably a corresponding "grid_label"  as well.

GEOSldas_App/ldas_setup
saraqzhang: we added 2 more GEOSldas_*.rc in GEOSldas_App/,  so "build "( build from adas or stand-alone ldas)  puts 6 of them in install*/etc/.    Setup will take the det and ens ones if it is coupled run,  if it is stand-alone the setup will just take the GEOSldas_hist.rc.

gmao-rreichle: Ok, thanks

GEOSmetforce_GridComp/LDAS_Forcing.F90
saraqzhang: I feel that replacing use_Predictor seems to be more disruptive to the current code.

GEOSldas_App/GEOSldas_HISTens.rc
gmao-rreichle: Just to clarify, for "lnd" and "lndfcstana", I'm only thinking of the **ensemble average** collections.  Not "lnd" for each ens member.

GEOSldas_App/sample_config_files/LADAS/exeinp.txt.Hy4dEnVar.atmens
```diff
@@ -11,14 +11,17 @@
...
+
+NUM_LDAS_ENSEMBLE: [NUM_ATMENS_ENSEMBLE]
 
 LADAS_COUPLING:    2
 
 ADAS_EXPDIR:       [full_path]/[ADAS_EXPDIR]
 
-MET_TAG:           [ADAS_EXPID]__Nx+- 
-MET_PATH:          [ADAS_EXPDIR]/atmens/ensdiag/mem
+MET_TAG:           [ADAS_EXPID]__bkg
```
gmao-rreichle, line 23: When we configure LDAS from within ADAS (fvsetup), ADAS_EXPID should be passed into ldas_setup via the new command-line arguments.  I assume (but don't know for sure) that we can then leave this blank in the exe_input file.  In any case, we need to add comments into this sample exe_input file that clarify which resource parameters need to be specified and which resource parameters are provided via fvsetup.

GEOSldas_App/sample_config_files/LADAS/exeinp.txt.Hy4dEnVar.central
```diff
@@ -11,12 +11,15 @@
...
+# e.g., CENTRAL_AGCM_GRID = CF0360x6C
+EXP_DOMAIN:        [CENTRAL_AGCM_GRID]_GLOBAL 
+
 LADAS_COUPLING:    1
 
 ADAS_EXPDIR:       [full_path]/[ADAS_EXPDIR]
 
-MET_TAG:           [ADAS_EXPID]__Nx+- 
-MET_PATH:          ../../scratch  
+MET_TAG:           [ADAS_EXPID]__bkg
```
gmao-rreichle, line 21: See comment above re. resource parameters provided via fvsetup

GEOSmetforce_GridComp/LDAS_Forcing.F90
gmao-rreichle: Good point.  The forcing in the LADAS is provided by the AGCM surface met forcing outputs.  As long as the corrected precipitation is handled correctly in the AGCM during the Predictor segment, the bkglfo files will already include the corrected precipitation as desired.  So there should **not** be a need to specify a MET_TAG with the corrected precipitation option when LDAS is running in the coupled atm/land DAS. 
We do need to make sure the corrected precipitation works as intended during the **Predictor** segment.  This should be the case, because we wouldn't want the ADAS bkg to have been created with model-generated precipitation.  But I'm not sure if that's ever been verified.

gmao-rreichle (commented):
@saraqzhang : I added a few inline comments, see below.

GEOSldas_App/sample_config_files/LADAS/exeinp.txt.Hy4dEnVar.atmens
saraqzhang: ADAS_EXPID and ADAS_EXPDIR are inputed from fvsetup to  ldas_setup, but I have not tested to leave the entries  blank in exeinp ( only tested  those set with some default strings ).  I will add a list of  parameters provided by fvsetup in the comments in the txt file, and it will be included in updating LADAS instruction on wiki.

saraqzhang:
I took a look at the Surface GC code,   if  precip correction is on and the data file available  for the period covering at least 12 hours starting from the rst time ,   the precip export (from Surface ) to lfo collection is the corrected quantity as long as AGCM runs the 12hr period, no dependency on  whether it is in "corrector" or " predictor".

GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -3252,6 +3255,46 @@ subroutine get_GEOS( date_time, force_dtstep, met_path, met_tag,             &
...
+    !
+    ! same as G5DAS except for file tag (column 3)
+
+    G5BKG_defs = G5DAS_defs
+
+    ! character(40):
+    !                            1         2         3         4
+    !                   1234567890123456789012345678901234567890     
+
+    G5BKG_defs( 1,3) = 'bkg.tavg1_2d_lfo                        '
```
gmao-rreichle, line 3271: The "bkg.lfo_*" Collection names here need to be consistent with whatever is finalized in https://github.com/GEOS-ESM/GEOSgcm_App/pull/707

gmao-rreichle (commented):
@saraqzhang: I tried to clean this up a bit.  In a nutshell, the "bkg.lfo_*" files replace what used to be the "Nx+-" files, so there's no need to keep both definitions, and keeping both would only confuse things.  Also, the intended logic is to determine "use_bkg" first, and then set GEOSgcm_defs accordingly.  Note that "use_bkg" is set based on the optional extensions to the MET_TAG (which are identified by the double underscore, see subroutine parse_G5DAS_met_tag()).  It should not be determined simply based on whether the string "bkg" appears in MET_TAG.   Please take a look and let me know if everything looks ok.  Thanks!

GEOSmetforce_GridComp/LDAS_Forcing.F90
gmao-rreichle: This piece of code was obsolete.  It was not doing anything.

saraqzhang:
I pulled the clean-up update and ran a test, passed.

saraqzhang:
@gmao-rreichle   modifications and additions are pushed  ( run tested )

GEOSldas_App/sample_config_files/LADAS/exeinp.txt.Hy4dEnVar.atmens
gmao-rreichle: @saraqzhang : were you able to test what happens when the rc variables that are provided by fvsetup are left blank in the LDAS.rc file?

gmao-rreichle:
@weiyuan-jiang : Please look at the changes in ldas_setup and let us know if the new code is ok with the overall design of the script.  Thanks!

GEOSldas_App/sample_config_files/LADAS/exeinp.txt.Hy4dEnVar.atmens
saraqzhang: I tested ADAS_EXPID left blank, it works, in ldas_setup it is a keyword "excluded from LDAS.rc".  But there are other variables indirectly rely on ADAS input, such as EXP_ID ( $ADAS_EXPID + '_LDAS' ),  it cannot be left blank, in ldas_setup  it is "Required exe input fields".

saraqzhang: correction:  tested ADAS_EXPDIR  left blank ( ADAS_EXPID is extracted from that) .

gmao-rreichle: Thanks, @saraqzhang.   It's probably best for now to keep the placeholders in the sample exeinp file.  At a later time, we can clean up what is and isn't required in the exeinp file to accommodate the new ldas_setup run mode (when fvsetup calls ldas_setup).  @weiyuan-jiang will look at the changes in ldas_setup to make sure they're consistent with the overall design of ldas_setup, which may (or may not) entail additional clean-up.

GEOSldas_App/sample_config_files/LADAS/exeinp.txt.Hy4dEnVar.central
```diff
@@ -7,16 +7,27 @@
...
+# (2) Use the resource parameter settings below when editing MY_exeinp.txt 
 #
+# (3) The following resource parameters are provided via fvsetup,
+#     which supersede the default in MY_exeinp.txt
+#      ADAS_EXPDIR
+#      ADAS_EXPID
+#      EXP_ID
+#      MET_TAG       
+#      BEG_DATE 
+#      GID
```
gmao-rreichle, line 19: @saraqzhang : I'm wondering about the following exeinp rc variables:
```
LADAS_COUPLING
MET_PATH
MET_HINTERP
LAND_ASSIM
LANDASSIM_DT, LANDASSIM_T0
FIRST_ENS_ID
ENSEMBLE_FORCING
JOB_SGMT, NUM_SGMT
```
Essentially, bullet (2) above says that the user needs to make sure the values of the rc variables in their exeinp file match the ones shown in the "sample" file below.  Since these values don't depend on anything that an LADAS user would change (assuming for now that the LADAS will always be run with 6-hour analysis windows), these values should be "forced" into ldas_setup by setup procedure.  That is, if a user specifies them incorrectly, fvsetup and ldas_setup still use the correct value.  

(1) One option to make this happen would be to add all of them to the list of command line args of ldas_setup, so the values could be hardwired into fvsetup.  But that's a lot of additional command line args.

(2) Since the values don't depend on the user's LADAS config (unlike, say, ADAS_EXPID or MET_TAG), we could expand ldas_setup to look for them in a "DEFAULT" file whenever ldas_setup is used in the LADAS config.  That is, we add two files to the source code (say, LADAS_exeinp_det.txt and LADAS_exeinp_ens.txt), and we hardcode the above rc variables into them.  Then we need to teach ldas_setup to look for these variables in the "LADAS_exeinp*.txt" file.  (We may need to have fvsetup furnish LADAS_COUPLING, so that ldas_setup knows whether to pick the "det" or "ens" version of the file.

(3) Or, alternatively, we could hardcode the values of these parameters directly into ldas_setup, and make sure that ldas_setup uses the hardcoded values whenever an LADAS experiment is configured. 

The three options differ only in the location of the hardcoded values.  In option (1), the hardcoded values live in fvsetup, in option (2) the hardcoded values live in "DEFAULT" files within the src code, and in option (3) the hardcoded values live in a ldas_setup [preferably in a separate function].

I think option (1) is least desirable because the hardcoded LDAS rc values would live in a distinct repo.  Perhaps option (3) would be best because it avoids creating additional src files. 

cc: @weiyuan-jiang

GEOSldas_App/sample_config_files/LADAS/exeinp.txt.Hy4dEnVar.central
```diff
@@ -7,16 +7,27 @@
...
+EXP_DOMAIN:        [CENTRAL_AGCM_GRID]_GLOBAL 
+
 LADAS_COUPLING:    1
 
-ADAS_EXPDIR:       [full_path]/[ADAS_EXPDIR]
+ADAS_EXPDIR:       [full_path]/[ADAS_EXPID]
 
-MET_TAG:           [ADAS_EXPID]__Nx+- 
-MET_PATH:          ../../scratch  
+MET_TAG:           [ADAS_EXPID]__bkg
```
gmao-rreichle, line 29: How do EXP_DOMAIN, ADAS_EXPDIR, and MET_TAG get assembled?  Is this done by fvsetup?  If so, we should add a comment that describes this. 
But, if we can implement the above suggestion (make ldas_setup use hardcoded rc variables for MET_HINTERP etc, do we even still need the present sample file?  Rather, it may make more sense to add a description of the process into a file called something like README_LADAS_setup

gmao-rreichle (commented):
@saraqzhang , see inline comments below

GEOSldas_App/sample_config_files/LADAS/exeinp.txt.Hy4dEnVar.central
saraqzhang: @gmao-rreichle  I agree, I like the option (3) best.  In the current PR version all things related to LADAS ( if coupled ) are scatted in ldas_setup, some info from ADAS, some info from exeinp example files.   If  we make a function in ldas_setup  (1) hold the default values and  (2) construct the parameters that depend on ADAS input ( currently done in the main code).   ldas_setup calls it to complete the setting special for coupled case.  
And we will no longer need  special coupled exeinp  example files in the repo.  
cc: @weiyuan-jiang

gmao-rreichle: Thanks, @saraqzhang, sounds good.  But we do need to keep the command-line inputs for EXP_ID etc because we can't predict the value and we don't want to have fvsetup write or edit a file that must then be read by ldas_setup.

GEOSldas_App/sample_config_files/LADAS/exeinp.txt.Hy4dEnVar.central
saraqzhang: ADAS_EXPDIR is a command line argument,  the others get assembled in ldas_setup. 
if we make a function in ldas_setup to handle all the default values and assembling, no need sample files.

gmao-rreichle: Ok, sounds good. But we still need to have a brief description of the handling of the various rc variables somewhere.

GEOSldas_App/sample_config_files/LADAS/exeinp.txt.Hy4dEnVar.central
saraqzhang: I will give it a try, adding a function and collect LADAS action items ( setting defaults and receiving fcsetup command-line inputs and assembling values)  into one call.  I may add more command line arguments in fvsetup for det or ens keyword, and resolutions.

gmao-rreichle: RESOLVING COMMENT FOR NOW SO PR CAN BE MERGED.

gmao-rreichle: REOPENING COMMENT AFTER MERGE TO MAKE IT EASIER TO IDENTIFY THE TASKS TO BE ADDRESSED IN A FUTURE PR.

PR #87, opened 2025-04-07
add python scripts for computing stats of ObsFcstAna output (via intermediate monthly files)

gmao-qliu:
Add python scripts for computing ObsFcstAna statistics.  

Processing is broken down into three steps:
1) Create intermediate monthly stats data (nc4).
2) Compute final stats from intermediate monthly data.
3) Aggregate stats across different species and/or remap stats data.

Steps 1) and 2) are generic for all ObsFcstAna output.  Stats are computed in the tile space of the DA experiment and for all obs species (separately) in the associated obs_param file. 

Step 3) depends on the assimilation experiment and use case scenario.  E.g., SMAP L4_SM stats are defined in the EASEv2 M09 tile space but are usually aggregated across H/V-pol and AM/PM ops and visualized on the EASEv2 M36 grid.  

The main script serves as a template to show how to use the pre-processed data to compute and plot the usual O-F and O-A statistical metrics. 

The intermediate monthly stats nc4 files contain:

**dimensions:**
```
tile    = [NUMBER_OF_TILES] ;
species = [NUMBER_OF_SPECIES] ;
```

**variables:**
```
char obs_param_assim(species) ;      

int N_data(tile, species) ;                ! number of obs

float obsxfcst_sum(tile, species) ;        ! sums of products of variables
float obsxana_sum(tile, species) ;
float fcstxana_sum(tile, species) ;

float obs_obs_sum(tile, species) ;         ! sums of variables
float obs_obsvar_sum(tile, species) ;
float obs_fcst_sum(tile, species) ;
float obs_fcstvar_sum(tile, species) ;
float obs_ana_sum(tile, species) ;
float obs_anavar_sum(tile, species) ;

float obs_obs2_sum(tile, species) ;        ! sums of squared variables
float obs_obsvar2_sum(tile, species) ;
float obs_fcst2_sum(tile, species) ;
float obs_fcstvar2_sum(tile, species) ;
float obs_ana2_sum(tile, species) ;
float obs_anavar2_sum(tile, species) ;
```

To facilitate the aggregation of the intermediate stats in Step 2), zeros are not replaced with NaNs in the "sum" variables even when the corresponding N_data value is zero. 

Related issue: https://github.com/GEOS-ESM/GEOSldas/issues/798 (closed)

### DIFF
--- a/GEOSldas_App/util/postproc/ObsFcstAna_stats/postproc_ObsFcstAna.py
+++ b/GEOSldas_App/util/postproc/ObsFcstAna_stats/postproc_ObsFcstAna.py
@@ -20,11 +20,11 @@ from helper.write_nc4            import write_sums_nc4, write_stats_nc4
 
 class postproc_ObsFcstAna:
     
-    def __init__(self, exp_list, start_time, end_time, sum_path='./', outid=None):
+    def __init__(self, exp_list, start_time, end_time, sum_path='./'):
         self.exp_list      = exp_list
         self.expdir_list   = [item['expdir'] for item in exp_list]
         self.expid_list    = [item['expid']  for item in exp_list]
-        self.exptag_list   = [item['exptag'] for item in exp_list]
+        self.exptag        = exp_list[0]['exptag']
         self.domain        = exp_list[0]['domain']
         self.start_time    = start_time 
         self.end_time      = end_time
@@ -45,11 +45,7 @@ class postproc_ObsFcstAna:
                 else:
                     self.obs_from = exp_idx
         if self.obs_from < 0: self.obs_from = 0       # by default, obs data are from exp_list[0]
-        print(f"obs data are from {self.exptag_list[self.obs_from]}")   
-
-        # Tag of the output sums files. If not provided, use main exptag 
-        if outid is None:
-            self.outid =  self.exptag_list[0]
+        print(f"obs data are from {exp_list[self.obs_from]['expid']}")   
 
         # Verify the configuration every time when current class is initialized
         # to avoid saving sums with different configs in the same directory
@@ -63,7 +59,7 @@ class postproc_ObsFcstAna:
             config_list.append({var:exp[var] for var in config_verify if var in exp})
 
         # File of configuration for verification
-        f_config = self.sum_path + '/' + self.outid + '_config.yaml'
+        f_config = self.sum_path + '/' + self.exptag + '_config.yaml'
 
         # Save a new file or compare current configuration with previously saved 
         if not os.path.exists(f_config):
@@ -194,7 +190,6 @@ class postproc_ObsFcstAna:
     def save_monthly_sums(self):
         expdir_list    = self.expdir_list
         expid_list     = self.expid_list
-        exptag_list    = self.exptag_list
         domain         = self.domain
         var_list       = self.var_list
         tc             = self.tilecoord
@@ -208,7 +203,7 @@ class postproc_ObsFcstAna:
             mo_path = self.sum_path + '/Y'+ date_time.strftime('%Y') + '/M' + date_time.strftime('%m') + '/'
             os.makedirs(mo_path, exist_ok=True)
     
-            fout = self.outid + '.ens_avg.ldas_ObsFcstAna_sums.' + date_time.strftime('%Y%m') +'.nc4'
+            fout = self.exptag + '.ens_avg.ldas_ObsFcstAna_sums.' + date_time.strftime('%Y%m') +'.nc4'
 
             fout = mo_path + fout
             
@@ -270,7 +265,7 @@ class postproc_ObsFcstAna:
                 
                 fpath = self.sum_path + '/Y'+ date_time.strftime('%Y') + '/M' + date_time.strftime('%m') + '/'
                     
-                fout  = self.outid + '.ens_avg.ldas_ObsFcstAna_sums.' + date_time.strftime('%Y%m') +'.nc4'
+                fout  = self.exptag + '.ens_avg.ldas_ObsFcstAna_sums.' + date_time.strftime('%Y%m') +'.nc4'
 
                 fout  = fpath + fout
                                 
@@ -355,7 +350,7 @@ class postproc_ObsFcstAna:
         var_list = ['obs_obs', 'obs_fcst','obs_ana']
         
         mo_path = self.sum_path + '/Y'+ date_time.strftime('%Y') + '/M' + date_time.strftime('%m') + '/'            
-        fnc4_sums = mo_path + self.outid + '.ens_avg.ldas_ObsFcstAna_sums.' + date_time.strftime('%Y%m') +'.nc4'
+        fnc4_sums = mo_path + self.exptag + '.ens_avg.ldas_ObsFcstAna_sums.' + date_time.strftime('%Y%m') +'.nc4'
         
         mdata_sum = {}
         mdata2_sum = {}
@@ -405,7 +400,7 @@ class postproc_ObsFcstAna:
             # var(x) = E[x2] - (E[x])^2
             data_var[var] = data2_mean[var] - data_mean[var]**2
 
-        # Computer metrics of O-F, O-A, etc. based on above stats
+        # Compute metrics of O-F, O-A, etc. based on above stats
         O_mean = data_mean['obs_obs']
         F_mean = data_mean['obs_fcst']
         A_mean = data_mean['obs_ana']
--- a/GEOSldas_App/util/postproc/ObsFcstAna_stats/user_config.py
+++ b/GEOSldas_App/util/postproc/ObsFcstAna_stats/user_config.py
@@ -26,13 +26,15 @@ def get_config():
 
     exp_main = {
         'expdir'        : '/discover/nobackup/projects/gmao/merra/iau/merra_land/SMAP_runs/SMAP_Nature_v11/',
-        'expid'         : 'DAv8_SMOSSMAP',
-        'exptag'        : 'DAMulti_SMAP', 
-        'domain'        : 'SMAP_EASEv2_M36_GLOBAL',
-        'da_t0'         : 3,                              # (fractional) UTC hour of first land analysis 
-        'da_dt'         : 10800,                          # ObsFcstAna file interval in seconds
-        'species_list'  : [5,6,7,8],                      # indices of species to be processed 
-        'obsparam_time' : "20150401_0000"                 # time stamp of obsparam file (YYYYMMDD_HHMM)
+        'expid'         : 'DAv8_SMOSSMAP',             # GEOSldas exp ID of simulation
+        'exptag'        : 'DAMulti_SMAP',              # string used in output file names for sums & stats,
+                                                       #   can be same as expid or different -- e.g., reflect info
+                                                       #   about subset of included species or about cross-masking
+        'domain'        : 'SMAP_EASEv2_M36_GLOBAL',  
+        'da_t0'         : 3,                           # (fractional) UTC hour of first land analysis 
+        'da_dt'         : 10800,                       # ObsFcstAna file interval in seconds
+        'species_list'  : [5,6,7,8],                   # indices of species to be processed 
+        'obsparam_time' : "20150401_0000"              # time stamp of obsparam file (YYYYMMDD_HHMM)
     }
 
     # Optional experiment(s) can be added for cross-masking or extracting obs from a different experiment.
@@ -40,7 +42,7 @@ def get_config():
     # All optional experiments and the main experiment must have identical tilecoords.
     # If the default "species" number/order do not match, set "species_list" accordingly to force a match.
     # Output will be cross-masked between all specified experiments.    
-
+    #
     # Forecasts and analyses are always from the main experiment.
     # Observations are from the experiment with 'use_obs' set to True (default is exp_main).  The most
     #   likely use case for reading obs from a supplemental experiment is when computing OmF etc diagnostics
@@ -50,21 +52,25 @@ def get_config():
     exp_sup1 = {
         'expdir'        : '/discover/nobackup/projects/gmao/merra/iau/merra_land/SMAP_runs/SMAP_Nature_v11/',
         'expid'         : 'DAv8_M36',
-        'exptag'        : 'DASMAP_SMAP', 
         'domain'        : 'SMAP_EASEv2_M36_GLOBAL',
-        'use_obs'       : True,                           # if True, use obs data from this exp
-        'species_list'  : [1,2,3,4],                      # indices of species to be processed 
-        'obsparam_time' : "20150401_0000"                 # time stamp of obsparam file (YYYYMMDD_HHMM)
+        'use_obs'       : True,                        # if True, use obs data from this exp
+        'species_list'  : [1,2,3,4],                   # indices of species to be processed,
+                                                       #   must identify same species as selected in main exp
+        'obsparam_time' : "20150401_0000"              # time stamp of obsparam file (YYYYMMDD_HHMM)
     }
 
-    # Convert experiments input to a list; first entry must be exp_main 
+    # Convert experiments input to a list; first entry must be exp_main: 
 
-    exp_list = [exp_main, exp_sup1]
+    #exp_list = [exp_main]               # no cross-masking
+    exp_list = [exp_main, exp_sup1]      # cross-mask exp_main with exp_sup1
     
-    # Top level directory to store monthly sum files; can use the experiment directory or a different path;
-    # /Yyyyy/Mmm/ is added automatically for individual months
+    # Top level directory for all output from this package:
 
-    out_path = '/discover/nobackup/qliu/SMAP_test/'
+    out_path = '/discover/nobackup/[USERNAME]/SMAP_test/'
+
+    # Directory for monthly sum files:
+    # - Can use the experiment directory or a different path.
+    # - Automatically appends /Yyyyy/Mmm/ for individual months.
     
     sum_path = out_path+'/'+exp_main['expid']+'/output/'+exp_main['domain']+'/ana/ens_avg/'
 
--- a/GEOSldas_App/util/postproc/ObsFcstAna_stats/Plot_stats_maps.py
+++ b/GEOSldas_App/util/postproc/ObsFcstAna_stats/Plot_stats_maps.py
@@ -2,6 +2,7 @@
 
 """
 Sample script for plotting maps of long-term data assimilation diagnostics.
+Computes Nobs-weighted avg of each metric across all species.
 Requires saved files with monthly sums (see Get_ObsFcstAna_stat.py).
 Stats of *normalized* OmFs are approximated!
 """
@@ -29,7 +30,7 @@ def plot_OmF_maps(postproc_obj, stats, fig_path='./'):
      
     Nmin = 20
 
-    # Then computer metrics of O-F, O-A, etc. based on above computed
+    # Then compute metrics of O-F, O-A, etc. based on above computed
     N_data = stats['N_data']
     O_mean = stats['obs_mean']
     # mean(x-y) = E[x] - E[y]   
@@ -66,7 +67,10 @@ def plot_OmF_maps(postproc_obj, stats, fig_path='./'):
     OmA_stdv[     N_data < Nmin] = np.nan
     N_data[       N_data < Nmin] = 0
 
-    # Combine metrics of individual species using weighted averaging
+    # Compute Nobs-weighted avg of each metric across all species.
+    # Typically used for SMAP Tb_h/h from asc and desc overpasses,
+    # or ASCAT soil moisture from Metop-A/B/C.
+    # DOES NOT MAKE SENSE IF, SAY, SPECIES HAVE DIFFERENT UNITS!
     Nobs_data     = np.nansum(              N_data, axis=1)
     OmF_mean      = np.nansum(OmF_mean     *N_data, axis=1)/Nobs_data
     OmF_stdv      = np.nansum(OmF_stdv     *N_data, axis=1)/Nobs_data
@@ -76,7 +80,7 @@ def plot_OmF_maps(postproc_obj, stats, fig_path='./'):
     OmA_stdv      = np.nansum(OmA_stdv     *N_data, axis=1)/Nobs_data
 
     # Plotting
-    expid = postproc_obj.outid
+    exptag = postproc_obj.exptag
 
     fig, axes = plt.subplots(2,2, figsize=(18,10))
     plt.rcParams.update({'font.size':14})
@@ -89,23 +93,23 @@ def plot_OmF_maps(postproc_obj, stats, fig_path='./'):
                 # crange is [cmin, cmax]
                 crange =[0, np.ceil((end_time-start_time).days/150)*300]
                 colormap = plt.get_cmap('jet',20)
-                title_txt = expid + ' Tb Nobs '+ start_time.strftime('%Y%m')+'_'+end_time.strftime('%Y%m')
+                title_txt = exptag + ' Tb Nobs '+ start_time.strftime('%Y%m')+'_'+end_time.strftime('%Y%m')
                 units = '[-]'
             if i == 0 and j ==1:
                 tile_data = OmF_mean
                 crange =[-3, 3]
                 colormap = plt.get_cmap('bwr', 15) 
-                title_txt = expid + ' Tb O-F mean '+ start_time.strftime('%Y%m')+'_'+end_time.strftime('%Y%m')
+                title_txt = exptag + ' Tb O-F mean '+ start_time.strftime('%Y%m')+'_'+end_time.strftime('%Y%m')
             if i == 1 and j == 0:
                 tile_data = OmF_stdv
                 crange =[0, 15]
                 colormap = plt.get_cmap ('jet',15)
-                title_txt = expid + ' Tb O-F stdv '+ start_time.strftime('%Y%m')+'_'+end_time.strftime('%Y%m')
+                title_txt = exptag + ' Tb O-F stdv '+ start_time.strftime('%Y%m')+'_'+end_time.strftime('%Y%m')
             if i == 1 and j == 1:
                 tile_data = OmF_norm_stdv
                 crange =[0, 15]
                 colormap = plt.get_cmap ('jet',15)
-                title_txt = expid + ' Tb normalized O-F stdv (approx!) '+ start_time.strftime('%Y%m%d')+'_'+end_time.strftime('%Y%m%d')
+                title_txt = exptag + ' Tb normalized O-F stdv (approx!) '+ start_time.strftime('%Y%m%d')+'_'+end_time.strftime('%Y%m%d')
 
             colormap.set_bad(color='0.9') # light grey, 0-black, 1-white
 
@@ -159,7 +163,7 @@ def plot_OmF_maps(postproc_obj, stats, fig_path='./'):
 
     plt.tight_layout()
     # Save figure to file
-    fig.savefig(fig_path+'Map_OmF_'+ expid +'_'+start_time.strftime('%Y%m')+'_'+\
+    fig.savefig(fig_path+'Map_OmF_'+ exptag +'_'+start_time.strftime('%Y%m')+'_'+\
                         (end_time+timedelta(days=-1)).strftime('%Y%m')+'.png')
     #plt.show()
     plt.close(fig)
--- a/GEOSldas_App/util/postproc/ObsFcstAna_stats/Plot_stats_timeseries.py
+++ b/GEOSldas_App/util/postproc/ObsFcstAna_stats/Plot_stats_timeseries.py
@@ -2,6 +2,7 @@
 
 """
 Sample script for plotting monthly time series of spatially averaged data assimilation diagnostics.
+Computes Nobs-weighted avg of each metric across all species.
 Requires saved files with monthly sums (see Get_ObsFcstAna_stat.py).
 """
 
@@ -17,11 +18,11 @@ from dateutil.relativedelta import relativedelta
     
 def Plot_monthly_OmF_bars(postproc_obj, fig_path='./'):
     import pickle
-    expid = postproc_obj.outid
+    exptag      = postproc_obj.exptag
     start_month = postproc_obj.start_time
-    end_month = postproc_obj.end_time
+    end_month   = postproc_obj.end_time
 
-    stats_file = fig_path + 'spatial_stats_'+expid+'_'+start_month.strftime('%Y%m')+ \
+    stats_file = fig_path + 'spatial_stats_'+exptag+'_'+start_month.strftime('%Y%m')+ \
             '_'+(end_month+timedelta(days=-1)).strftime('%Y%m')+'.pkl'
     
     if os.path.isfile(stats_file):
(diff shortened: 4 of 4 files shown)


GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/compute_monthly_stats.py
```diff
@@ -0,0 +1,87 @@
+import numpy as np
+from datetime import datetime, timedelta
+from dateutil.relativedelta import relativedelta
+from helper.read_GEOSldas import read_ObsFcstAna, read_tilecoord, read_obs_param
+
+def compute_monthly_stats(expdir,expid,domain,this_month,tc,obs_param,var_list):
```
gmao-rreichle, line 6: Assuming this is "Step 1)" in the processing (per the introductory comment of the PR), it would probably be better to move this script up one directory level.  Or maybe provide a (very short) driver script that executes this step for a user-specified period?   Trying to separate the "generic" Steps 1) and 2) from the final Step 3) that is very dependent on the experiment and use case.

GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/plot.py
```diff
@@ -0,0 +1,68 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from mpl_toolkits import basemap
+
+def plotMap(
```
gmao-rreichle, line 5: This is extremely generic and should be moved to ./util/shared/python if possible.

GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/write_nc4.py
```diff
@@ -0,0 +1,36 @@
+from netCDF4 import Dataset
+import numpy as np
+
+def write_sums_nc4(file_path, N_data, data_sum, data2_sum, oxf_sum, oxa_sum, fxa_sum, obs_param):
```
gmao-rreichle, line 4: I suggest renaming the file to "write_sums_nc4.py" for consistency with the py function contained in the file.  (When renaming the file, use "git mv old name new name" so that git can keep track of the renaming.)

GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/read_GEOSldas.py
```diff
@@ -0,0 +1,224 @@
+import numpy as np
```
gmao-rreichle, line 1: Move to ./util/shared/python if possible.  (When moving/renaming the file, use "git mv olddir/oldname newdir/newname" so that git can keep track of the renaming.)

GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/util.py
```diff
@@ -0,0 +1,29 @@
+import os
```
gmao-rreichle, line 1: Move to ./util/shared/python if possible.  (When renaming the file, use "git mv oldname newname" so that git can keep track of the renaming.)

GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/util.py
```diff
@@ -0,0 +1,29 @@
...
+import numpy as np
+
+def make_folder(folder_path):
+    if not os.path.exists(folder_path):
+        os.makedirs(folder_path)
+
+def crd2grid(y, x):
```
gmao-rreichle, line 8: needs plain English statement of what the function is doing

GEOSldas_App/util/postproc/ObsFcstAna_stats/main_ObsFcstAna.py
```diff
@@ -0,0 +1,266 @@
...
+
+"""
+GEOSldas DA Diagnostics Map Generator
+
+This script computes and plots GEOSldas Data Assimilation (DA) diagnostics maps
+based on the FcstObsAna output. It processes monthly data and generates final
+metrics from these monthly statistics.
+
+The script performs the following main tasks:
+1. Loads and processes monthly FcstObsAna output data
```
gmao-rreichle, line 11: I'm wondering if it makes sense to split this "main" script into two or three pieces. The first piece would deal with creating the monthly intermediate stats (Step 1) for a user-specified date range.
The second piece would compute the stats for the same date range from the monthly intermediate files (Step 2). 
(Or maybe Steps 1 and 2 can be combined into one script, which is essentially the first portion of the present "main" script.).  The main idea here is that the user should never need to edit the Step1/2 scripts (except for providing the info about the experiment name & location and the desired stats period).   The result of Step1/2 could be another nc4 file, or maybe the Step 1/2 script could be called in just a single line at the beginning of the Step 3 (template/"main") script. 
The third piece would then deal with the final aggregation (across species) and remapping and the plotting, all of which which is very dependent on the use case.

gmao-rreichle (commented):
@gmao-qliu, I added a few thoughts.  Please take a look and let me know if you have any questions or concerns.  Thanks!

GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/smapeasev2.py
```diff
@@ -0,0 +1,214 @@
...
+    #  REFERENCE         Brodzik, M. J., B. Billingsley, T. Haran, B. Raup, and M. H.
+#                    Savoie (2012): EASE-Grid 2.0: Incremental but Significant
+#                    Improvements for Earth-Gridded Data Sets. ISPRS International
+#                    Journal of Geo-Information, vol. 1, no. 1, pp. 32-45,
+#                    http://www.mdpi.com/2220-9964/1/1/32/
+#  
+#  Steven Chan, 11/2011
+#  Email: steven.k.chan@jpl.nasa.gov
+    
+def smapeasev2_ind2latlon(row=None,col=None,gridid=None,*args,**kwargs):
```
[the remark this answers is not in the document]
gmao-qliu: > For consistency with GEOSldas, this subroutine should be called EASEv2_ind2latlon(). The function is only about the EASE grid and has nothing to do with "smap" (except that SMAP happens to use the EASE grid). Also, ideally this function would be in a different folder. We keep the matching Matlab helper script in ./util/shared/matlab, so the python script should be in ./util/shared/python. I don't know how this changes the "import" statements in the main py script, but I hope this can be accommodated somehow. (When renaming the file, use "git mv oldname newname" so that git can keep track of the renaming.)

I create util/shared/python and moved/will move some codes there. We can use the following to add the directory in the search path for import. It works like "addpath" in matlab.  

import sys
sys.path.append('../../shared/python/')

GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/compute_monthly_stats.py
gmao-qliu: I moved compute_monthly_stats() out of helper/. It can be used as a stand-alone script. The "driver" is toward the end after the line 
if __name__ == '__main__':

gmao-rreichle:
@gmao-qliu : I removed the binary files in the __pycache__/ directories from the repository.  I don't think these should be under version control.

GEOSldas_App/util/postproc/ObsFcstAna_stats/main_ObsFcstAna.py
gmao-qliu: I'm not sure about this.The code for Step 1 already contains specific assumptions (such as 3-hour timesteps and monthly summations requiring complete month data). I still think users should consider these scripts as templates rather than black box solutions.

gmao-rreichle: The output time step and the requirement to have a complete month isn't any different from what the post-processing of the monthly avg HISTORY needs.  Note that we also hardwire "dt_assim" (a.k.a. "dtstep_assim" in GEOSldas) in the util routine that computes the scaling parameters https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/9d304a0ac47b173ee8b0b027a88be87a75a2f5aa/GEOSldas_App/util/inputs/obs_scaling_params/Run_get_model_and_obs_clim_stats_latlon_grid.m#L44
I forget where dtstep_assim is defined in the config.  If it's not possible to extract it from a file in the rc_out dir, then we could just require it as an input to the python postproc routine.  It's a property of the experiment and in this sense very generic (just like the exp name and dir).   The processing routine could then also check for gaps in the ObsFcstAna time series and exit with a message if there is a gap

GEOSldas_App/util/shared/python/read_GEOSldas.py
```diff
@@ -0,0 +1,224 @@
...
+            param['bias_Npar'] = float(fid.readline().strip())
+            param['bias_trel'] = float(fid.readline().strip())
+            param['bias_tcut'] = float(fid.readline().strip())
+            param['nodata'] = float(fid.readline().strip())
+            param['varname'] = fid.readline().strip().strip('"')
+            param['units'] = fid.readline().strip().strip('"')
+            param['path'] = fid.readline().strip().strip('"')
+            param['name'] = fid.readline().strip().strip('"')
+
+            if 'Vv80' in fname or 'OL80' in fname:
```
amfox37, line 39: Not sure about this - I guess it is for backwards compatibility? I think any experiments run with the current release or later will be outputting maskpath and maskname here, so I've been using:
            param['maskpath'] = fid.readline().strip().strip('"')
            param['maskname'] = fid.readline().strip().strip('"')

gmao-qliu: Right. The old file doesn't have 'maskpath' and 'maskname' and thus contains two fewer lines. The current code definitely needs some modification. I haven't given much thought to it.

GEOSldas_App/util/postproc/ObsFcstAna_stats/main_ObsFcstAna.py
gmao-qliu: I updated the script by using an ObsFcstAna_prep class. Step 1 & Step 2 are now functions of the class object. There is no time loop in the 'main' script.

GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/compute_monthly_stats.py
```diff
@@ -0,0 +1,105 @@
...
+                    if obs_param[ispec]['assim'] == 'T':
+                        is_valid = np.logical_and(OFA['obs_species'] == this_species, OFA['obs_assim']==1)
+                    else:
+                        is_valid = OFA['obs_species'] == this_species
+
+                    masked_tilenum = OFA['obs_tilenum'][is_valid]
+                    for var in var_list:
+                        masked_data[var] = OFA[var][is_valid]
+          
+                    tile_idx = np.where(np.isin(tc['tile_id'], masked_tilenum))[0]
```
amfox37, line 71: I think we need something like the following to deal with regional runs when `OFA['obs_tilenum']` and `tc['tile_id']` will no longer align:

 ```
                # Convert to 0-based indices for tc array
                tile_indices = masked_tilenum - 1
        
                # Get corresponding tc tile IDs
                tc_tile_ids = tc['tile_id'][tile_indices]

                tile_idx = np.where(np.isin(tc['tile_id'], tc_tile_ids))[0]
```

Comment (gmao-qliu) [reply GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/compute_monthly_stats.py:71]: Seem we don't need 'tile_id' here, does it make sense to change L67 to tile_idx = OFA['obs_tilenum'][is_valid], and remove L71?

Comment (gmao-qliu) [reply GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/compute_monthly_stats.py:71]: correction,      tile_idx = OFA['obs_tilenum'][is_valid] - 1 to make it 0-based

Comment (amfox37) [reply GEOSldas_App/util/postproc/ObsFcstAna_stats/helper/compute_monthly_stats.py:71]: Yes, I think `tile_idx = OFA['obs_tilenum'][is_valid] - 1 ` is correct. The actual 'tile_id' isn't required here.

**Comment by amfox37** [comment]
I have gone ahead and added the capability to calculate monthly OmF mean etc statistics from the precomputed sums/sum of squares.  I'm not entirely sure it belongs in the Main_example.py as I've inserted it, let's discuss. 
Also, have only run a very limited test case locally, so will now do some more substantial testing on Discover.

File: GEOSldas_App/util/postproc/ObsFcstAna_stats/Main_example.py [hunk 11]
```diff
@@ -0,0 +1,407 @@
...
+
+# Define time range for processing (Year, Month, Day)
+start_time = datetime(2015,4,1)
+end_time   = datetime(2016,4,1)
+
+# -------------------------------- Experiments Information -----------------------------------------
+# Supports single experiment or multiple experiments.
+# All experiments must have identical tilecoords and number/order of observation species.
+# If the default "species" number/order do not match, need to set the *optional*
+#   "select_species" key to get a match, i.e. same species sequences.
```
gmao-rreichle, line 56: I'm a bit confused by these two seemingly contradictory comments.  The 2nd comment seems to suggest it is possible for multiple experiments to **not** have the same number/order of obs species

GEOSldas_App/util/postproc/ObsFcstAna_stats/Main_example.py
```diff
@@ -0,0 +1,407 @@
...
+OmA_stdv[     N_data < Nmin] = np.nan
+
+# Combine metrics of individual species using weighted averaging
+OmF_mean      = np.nansum(OmF_mean     *N_data, axis=1)/np.nansum(N_data,axis=1)
+OmF_stdv      = np.nansum(OmF_stdv     *N_data, axis=1)/np.nansum(N_data,axis=1)
+OmF_norm_mean = np.nansum(OmF_norm_mean*N_data, axis=1)/np.nansum(N_data,axis=1)
+OmF_norm_stdv = np.nansum(OmF_norm_stdv*N_data, axis=1)/np.nansum(N_data,axis=1)
+OmA_mean      = np.nansum(OmA_mean     *N_data, axis=1)/np.nansum(N_data,axis=1)
+OmA_stdv      = np.nansum(OmA_stdv     *N_data, axis=1)/np.nansum(N_data,axis=1)
+Nobs_data     = np.nansum(              N_data, axis=1)
```
gmao-rreichle, line 180: Can the computation of Nobs_data be moved up, so it can be used for normalizing, rather than computing  "np.nansum(N_data,axis=1)" in each of the preceding 6 lines?

GEOSldas_App/util/postproc/ObsFcstAna_stats/Main_example.py
```diff
@@ -0,0 +1,407 @@
...
+from mpl_toolkits.basemap   import Basemap
+
+from read_GEOSldas          import read_tilecoord, read_obs_param
+from util                   import make_folder, array2grid
+from plot                   import plotMap
+from EASEv2                 import EASEv2_ind2latlon
+
+from postproc_ObsFcstAna    import postproc_ObsFcstAna
+
+# Uncomment if to run the script in the background to see the standard output while running
```
gmao-rreichle, line 43: I don't understand this comment.

GEOSldas_App/util/postproc/ObsFcstAna_stats/postproc_ObsFcstAna.py
```diff
@@ -0,0 +1,537 @@
...
+        expid_list    = self.expid_list
+        exptag_list   = self.exptag_list
+        domain        = self.domain
+        var_list      = self.var_list
+        tc            = self.tilecoord
+        obsparam_list = self.obsparam_list
+        start_time    = self.start_time
+        end_time      = self.end_time
+
+        start_month =  start_time.replace(day=1,hour=3)
```
gmao-rreichle, line 48: LANDASSIM_DT and LANDASSIM_T0 should not be hardwired here and in a couple of other places throughout the scripts.  They can be specified by the user as variables in the Main script.   (That is, no need to extract them from the experiment directory.)

gmao-rreichle (commented):
@gmao-qliu, @amfox37:  

- I went through the scripts and added a couple of commits with cleanup changes: https://github.com/GEOS-ESM/GEOSldas_GridComp/compare/e3d2339..ef52a11?w=1  (white-space changes hidden)
- I also added a few inline comments, see below. 
- Re. Andy's changes of adding monthly stats calcs (https://github.com/GEOS-ESM/GEOSldas_GridComp/compare/c951034..e3d2339), I feel we need to keep working on the modularity and nomenclature. 
  - I'm now thinking of this package having two major use cases: (1) creating maps of long-term DA diagnostics and (2) creating time series of spatially averaged DA diagnostics. I assume Andy's addition is targeting the latter, but I'm not entirely sure.
  - In addition to the main use cases, we also have the necessary intermediate step of computing the monthly (intermediate) files with sums, which is shared across the two use cases.
  - Perhaps we should split the current Main_example.py into three separate scripts: (a) for a specified period, compute monthly (intermediate) files with sums, (b) create (and plot?) maps (2d-arrays) of long-term DA diagnostics, and (c) create (and plot?) monthly time series of spatially averaged DA diagnostics.  For the latter, I'm not sure we want to save monthly maps of DA diagnostics -- by themselves, they're not very meaningful. 
  - For the time series of months DA diagnostics, we need to agree on the exact algorithm to compute the monthly and spatially averaged stats. My take here is that it's probably best to use equal weights for all individual OmF values.  That is, do not compute the monthly DA diagnostics first and then take a spatial average.  Rather, compute the DA diagnostics from all (pooled) OmF values that are available for the month of interest (regardless of location).  Not sure what's currently implemented in Andy's functions.  
- The file names `write_nc4.py`, `plot.py`, and `util.py` are very generic.  We may want to give them somewhat more specific names.   The latter two are also missing comments about what they're meant to do. 
- Throughout the package, we should be consistent in the terminology re. the (intermediate) files with "sums" and the final files with "stats" (or DA diagnostics/metrics).  At least I think that was the overall intent, and I renamed the compute_monthly_stats.py script to compute_monthly_sum.py to make this more consistent.  I hope I didn't mess this up.

GEOSldas_App/util/postproc/ObsFcstAna_stats/Main_example.py
```diff
@@ -0,0 +1,407 @@
...
+from postproc_ObsFcstAna    import postproc_ObsFcstAna
+
+# Uncomment if to run the script in the background to see the standard output while running 
+# import io
+#sys.stdout = io.TextIOWrapper(open(sys.stdout.fileno(), 'wb', 0), write_through=True)
+#sys.stderr = io.TextIOWrapper(open(sys.stderr.fileno(), 'wb', 0), write_through=True)
+
+# Define time range for processing (Year, Month, Day)
+start_time = datetime(2015,4,1)
+end_time   = datetime(2016,4,1)
```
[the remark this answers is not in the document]
gmao-qliu: datetime() needs the "day" input. Also, the current script can handle partial month: it'll save a sum file with slightly different filename depending on the "day" value in the start_time and end_time.

GEOSldas_App/util/postproc/ObsFcstAna_stats/Main_example.py
gmao-qliu: Without these lines, the redirected standard output won't be seen in the out.log file until the job is done if we run the script in the background, e.g., Main_example.py > out.log &. We don't need it, but it's convenient to have for debugging purposes. They need to be commented out in a foreground run.

GEOSldas_App/util/postproc/ObsFcstAna_stats/Main_example.py
gmao-qliu: experiments don't need to have the same species sequences, but the select_species should result in the same species sequences. I will think of better wording, but feel free to edit.

GEOSldas_App/util/postproc/ObsFcstAna_stats/postproc_ObsFcstAna.py
gmao-qliu: They are now specified by user. I removed the hard coded part.

amfox37: The Plot_stats_maps.py example script is now missing this specification. i.e missing:
          'da_t0'        : 3,       # first hour of the month 
          'da_dt'        : 10800,   # ObsFcstAna file interval in seconds

gmao-rreichle:
@gmao-qliu : Thanks for reorganizing the Main script.  I tried to clean up Save_monthlysums.py by more cleanly separating the user-defined inputs and the generic processing: https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/87/commits/6165af823cff975d325e0c61e82894936167bd2b
I did not test the code, so chances are it won't run without further debugging.  For now, please take a look when you get a chance and see if the changes make sense to you. 
What I don't quite understand yet is whether sums files will be stored for both "exp_main" and "exp_sup1", or whether sums files are stored only for "exp_main".  In other words, if I have two experiments for which I would like sums files, do I have to run Save_monthlysums.py once or twice?  Either way is fine, this info just needs to be clarified in the comments section. 
I'll take a look at the other scripts once I know if any of my proposed changes to Save_monthlysums.py make sense.

gmao-qliu:
I tested the new changes and updated with a few minor. Everything works for me.

GEOSldas_App/util/postproc/ObsFcstAna_stats/user_config.py
```diff
@@ -0,0 +1,132 @@
...
+    # Convert experiments input to a list; first entry must be exp_main 
+
+    exp_list = [exp_main, exp_sup1]
+    
+    # Forecasts and analyses are always from the main experiment.
+    # Observations can be from experiment indicated by 'obs_from' index (0-based).
+    # The mostly likely use case for this is that _scaled_ observations from a DA experiment
+    #   are used to compute OmF etc diagnostics for a corresponding open loop experiment.
+
+    obs_from = 1              # 0-based index of exp in exp_list from which obs are extracted
```
gmao-rreichle, line 62: Just a thought -- would it be cleaner to have a new field "use_obs" as part of the exp_XXXX structure?  That would get rid of the additional variable.  We can add a check that exactly one exp in exp_list has use_obs=true, all others must be false.

GEOSldas_App/util/postproc/ObsFcstAna_stats/user_config.py
```diff
@@ -0,0 +1,132 @@
...
+    # Sums or stats will be processed for exp_main:
+
+    exp_main = {
+        'expdir'        : '/discover/nobackup/projects/gmao/merra/iau/merra_land/SMAP_runs/SMAP_Nature_v11/',
+        'expid'         : 'DAv8_SMOSSMAP',
+        'exptag'        : 'DAMulti_SMAP', 
+        'domain'        : 'SMAP_EASEv2_M36_GLOBAL',
+        'da_t0'         : 3,                              # hour of first land analysis
+        'da_dt'         : 10800,                          # ObsFcstAna file interval in seconds
+        'species_list'  : [5,6,7,8],
```
gmao-rreichle, line 34: I'm not sure I fully understand "species_list".  Is this an index vector into the species defined in "obsparam"?  That is, from exp_main, we want the stats for the 5th, 6th, 7th, and 8th species within the "obsparam" vector?  I forget if we enforce that the species ID in the obsparam vector is always a consecutive number from 1 to N_obsparam.  If that's the case, the indices here match the species "ID" in the obsparam.  But I'm not sure this is generally the case.  So further tweaking of the meaning of "species_list" is probably required in the py script.

GEOSldas_App/util/postproc/ObsFcstAna_stats/user_config.py
```diff
@@ -0,0 +1,132 @@
...
+        
+        fop      = expdir+expid+'/output/'+domain+'/rc_out/Y'+YYYY+'/M'+MM+'/'+expid+'.ldas_obsparam.'+exp['obsparam_time']+'z.txt'
+        obsparam = read_obs_param(fop)
+
+        # get the species list and default to list of all species if doesn't exist 
+        species_list = exp.get('species_list',[int(obsparam[i]['species']) for i in range(len(obsparam))])
+        
+        # reorder obsparam to match across experiments
+        obsparam_new = []
+        for i in range(len(obsparam)):
```
gmao-rreichle, line 111: Maybe here or in a separate loop we can assert that the obsparam info across the experiments does indeed match (detect faulty user input)

GEOSldas_App/util/shared/python/read_GEOSldas.py
```diff
@@ -0,0 +1,239 @@
...
+# ----------------------------------------------------------------------------
+#
+# reader for GEOSldas tilecoord file (binary)
+
+def read_tilecoord(fname):
+    int_precision   = 'i'
+    float_precision = 'f'
+
+    # SPECIFY endianness
+    machfmt = '<'                    # '>' for big-endian, '<' for little-endian
```
gmao-rreichle, line 75: could/should this be moved to the top of the "module" (global variable for the module, in Fortran speak)?  It's rather hidden in the middle of things here. Note that the same occurs in the next function.

GEOSldas_App/util/postproc/ObsFcstAna_stats/postproc_ObsFcstAna.py
```diff
@@ -0,0 +1,392 @@
...
+            self.outid =  '_'.join([item for item in self.exptag_list])
+
+            if self.obs_from  > 0:
+                self.outid = self.outid + '_Obs_from_' + self.exptag_list[self.obs_from]
+    
+    # ----------------------------------------------------------------------------------------------------------
+    #
+    # Function to compute monthly sums from x-hourly ObsFcstAna data for a single month.
+
+    def compute_monthly_sums(self, date_time):
```
gmao-rreichle, line 49: I replaced the variable names *month with date_time for clarity.  If I understood correctly, the variable "month" included the full year/month/day/hour/min/sec info, so calling it "month" was very confusing.  I applied the same changes in several functions.  Please double-check carefully.  I hope I didn't mess it up.

GEOSldas_App/util/postproc/ObsFcstAna_stats/postproc_ObsFcstAna.py
```diff
@@ -0,0 +1,392 @@
...
+            
+            # make monthly file output directory 
+            mo_path = self.sum_path + '/Y'+ date_time.strftime('%Y') + '/M' + date_time.strftime('%m') + '/'
+            os.makedirs(mo_path, exist_ok=True)
+    
+            fout = self.outid + '.ens_avg.ldas_ObsFcstAna.' + date_time.strftime('%Y%m') +'_sums.nc4'
+
+            fout = mo_path + fout
+            
+            # skip if output file already exists
```
gmao-rreichle, line 179: I ran out of time to look into this in more detail, but are we ensuring that an existing sums file is consistent with new sums files generated for other months in terms of exp_list (i.e., cross-masking and "obs_from")?

GEOSldas_App/util/postproc/ObsFcstAna_stats/Save_monthlysums.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # Initialize the postprocessing object
+    postproc = postproc_ObsFcstAna(exp_list, start_time, end_time, obs_from=obs_from, sum_path=sum_path)
+
+    # Compute and save monthly sums
+    postproc.save_monthly_sums()
+
+    # --------------------------------------------------------------------------------------
+    # Optionally compute long-term temporal/spatial statistics and create plots.
+    # The plotting scripts can also run standalone using the individual Plot_stats_*.py scripts,
```
gmao-rreichle, line 67: I like the optional plotting below.  I don't think we need to include main() functions in the Plot_stats*.py files.  Perhaps the best way to organize the code is to create a single "plot_OmF_stats.py" that contains the functions defined in the current Plot_stats*.py scripts.  Perhaps we can add a switch that lets the user skip the "save_monthly_sums()" action, in case all they want to do is plot.  Does that make sense?

GEOSldas_App/util/postproc/ObsFcstAna_stats/user_config.py
gmao-qliu: suggestion is implemented.

GEOSldas_App/util/postproc/ObsFcstAna_stats/user_config.py
gmao-qliu: This is indeed the case as far as I'm aware. For example, we have obsparam IDs 1-4 for SMOS species and 5-8 for SMAP species in DA_SMOSSMAP, and 1-4 for SMAP species in DA_SMAP. The order is based on the ensupd input. Maybe @amfox37 can confirm.

GEOSldas_App/util/shared/python/read_GEOSldas.py
gmao-qliu: I am not sure if we can do that without making a class. Let's keep it as is for now.

GEOSldas_App/util/postproc/ObsFcstAna_stats/postproc_ObsFcstAna.py
gmao-rreichle: Ok, thanks.  This should work if we have no more than 2 or maybe 3 experiments in exp_list, and if the exp names are relatively short.  Otherwise the file names of the sums files might become too long.  Let's see if what we have now works for Andy.  We might still need to add the "config" to each nc4 file.  This should make it straightforward to read "config" from an existing monthly sums file and verify it against the "config" of the current run.

gmao-rreichle (commented):
@gmao-qliu : I added a few more commits:
1. Here's a link that shows the cumulative changes (white-space differences are hidden): https://github.com/GEOS-ESM/GEOSldas_GridComp/compare/f02efb4..fe2b15a?w=1
The most likely source of errors in my changes is in my addition of a check for obsparam across the exp_list:  https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/87/commits/d1852a8153b2472072567ab926251d40fdd1d5d1
Please take a look and let me know if anything needs fixing.  
2. What did we decide to do about the stats of the normalized O-F?  If I recall correctly, as implemented the stats differ from what we did with the old Matlab scripts.  Are we still thinking of trying to change this so it becomes identical to what we did in Matlab?  If not, we need to add documentation in Plot_stats_maps.py.

gmao-qliu:
> @gmao-qliu : I made a few more changes to clean up and clarify things. Here's a link to the cumulative differences of my commits of today: https://github.com/GEOS-ESM/GEOSldas_GridComp/compare/84c3b58..5d42392
> 
> This leaves the following comment as the only one that may still need resolution: [#87 (comment)](https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/87#discussion_r2121895300) (re. the potentially excessive length of the names of the files that hold the monthly sums)

@gmao-rreichle, Please review my latest commit to address this issue.

gmao-rreichle:
@gmao-qliu, I added one more commit with clean-up: https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/87/commits/990bce2a94dd148f9219465ec30557a55714d4f8.  
Besides adding new and clarifying existing comments, I simplified the variables related to exptag, exptag_list, and outid.  I think we only need one string for all of this, which I called "exptag".  The others seem to be superfluous.  Please double check and let me know, thanks!

gmao-qliu:
> @gmao-qliu, I added one more commit with clean-up: [990bce2](https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/990bce2a94dd148f9219465ec30557a55714d4f8). Besides adding new and clarifying existing comments, I simplified the variables related to exptag, exptag_list, and outid. I think we only need one string for all of this, which I called "exptag". The others seem to be superfluous. Please double check and let me know, thanks!
@gmao-rreichle Looks and tests all good to me.

gmao-rreichle (approved):

PR #94, opened 2025-04-25
additional default settings and command line args for land-atm DAS

saraqzhang:
Updates ldas_setup for improved integration of land-atm DAS config. 

New functionality needed for revisions in https://github.com/GEOS-ESM/GEOSadas/pull/352 (for coupled land-atm experiments).

Successfully 0-diff tested by @gmao-rreichle after https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/3d9121e9d80e8bf4efa6fa254bb367fc3663d55f

---

For the record: Spreadsheet with GEOSldas rc variables as discussed after https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/ab9c3c39a7c875102d1cc30323f25ac15cda3e63 (~7-8 May 2025)
[LADAS_exeinp_variables_20250508T1113.xlsx](https://github.com/user-attachments/files/20416045/LADAS_exeinp_variables_20250508T1113.xlsx)

---

Left for **future cleanup** of ldas_setup:
- Identify list of truly "required" inputs and adjust py variable names in ldas_setup accordingly (https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94#discussion_r2071675590).
- Create function for processing GEOS_SurfaceGridComp.rc (https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94#issuecomment-2902744581)
- Can we have ldas_setup print the main help message (i.e., the help message that is printed for `ldas_setup -h`) when run without any arguments (i.e., just `ldas_setup`, which currently exits with an error)?  
- Should we rename ldas_setup to ldas_setup.py to identify it as a py script? (Not sure if that's common practice or not.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added more default settings and command line arguments for integrated land-atm DAS setup 
+- Added functionality for model-based QC of MODIS SCF observations using soil layer 1 temperature.
 - Added functionality to simulate landice tiles.
 - Added functionality to read nc4-formatted tile file.
 
--- a/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
+++ b/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
@@ -1213,8 +1213,9 @@ contains
           
        case('asnow')
 
-         get_asnow_l   = .true.
-         get_asnow_lH  = .true.
+          get_asnow_l   = .true.
+          get_asnow_lH  = .true.
+          get_tp_l      = .true.   ! needed for model-based QC
          
        case default
           
@@ -1365,7 +1366,7 @@ contains
 
           ! updated to new interface - reichle, 3 Apr 2012
           
-          call catch_calc_tp( N_catl, cat_param%poros,                  &
+          call catch_calc_tp( N_catl, cat_param%poros,                           &
                catprogn2ghtcnt(N_catl,cat_progn(:,n_e)), tp_l )
                     
        end if
@@ -1461,8 +1462,11 @@ contains
                 call qc_model_based_for_Tb( N_catl, precip, Tb_v_l(:,j,n_e) )
                 
              end do
-             
+
           end if
+
+          if (get_asnow_lH) &
+               call qc_model_based_for_asnow( N_catl, tp_l(1,:), asnow_l(:,n_e) )   
 
        end if
 
@@ -2254,7 +2258,7 @@ contains
 
     do i=1,N_cat
        
-       ! delete obs 
+       ! delete Obs_pred 
        ! - if there is snow on the ground 
        ! - if it is raining/snowing
        ! - if surface temperature is around or below freezing
@@ -2337,7 +2341,7 @@ contains
     
     do i=1,N_cat
        
-       ! delete obs 
+       ! delete Obs_pred 
        ! - if there is snow on the ground 
        ! - if it is raining/snowing
        ! - if "avoid_frozen" and frozen
@@ -2391,7 +2395,7 @@ contains
 
     do i=1,N_cat
        
-       ! delete obs 
+       ! delete Obs_pred 
        ! - if there is heavy rain or snow
        
        ! NOTE: subroutine mwRTM_get_Tb already returns no-data-values
@@ -2412,6 +2416,44 @@ contains
     end do
     
   end subroutine qc_model_based_for_Tb
+
+    ! *****************************************************************
+  
+  subroutine qc_model_based_for_asnow( N_cat, tp1, asnow )
+    
+   ! Model-based quality control for MODIS SCF observations.
+   ! Sets "asnow" to no-data when layer-1 soil temperature exceeds a threshold.
+   !
+   ! amfox, 7 May 2025
+   !
+   ! --------------------------------------------------------------
+   
+   implicit none
+   
+   integer,                   intent(in)    :: N_cat
+   
+   real,    dimension(N_cat), intent(in)    :: tp1              ! layer-1 soil temperature [C]
+   
+   real,    dimension(N_cat), intent(inout) :: asnow            ! snow cover fraction [0-1]    
+   
+   ! local variables
+   
+   real, parameter :: temperature_threshold = 10.               ! [C]
+   
+   integer :: i
+   
+   ! ---------------------------------------    
+
+   do i=1,N_cat
+      
+      ! delete Obs_pred 
+      ! - if the layer-1 soil temperature exceeds threshold
+      
+      if (tp1(i) > temperature_threshold)  asnow(i) = nodata_generic
+
+   end do
+   
+ end subroutine qc_model_based_for_asnow
   
   ! *********************************************************************
 


saraqzhang:
@gmao-rreichle  This PR  contains the addition/modification to ldas_setup on top of ldasGC 2.0.0.   When the coupling is on, it sets default values to a list of  parameters. It also accepts a few more command line options that allow fvsetup pass adas information for the coupling.  With these updates  fvsetup will query where to get  ldas initial conditions and pass them to ldas_setup.  We only need  one exeinp.txt ( work for both det and ens) .   The modifications in fvsetup corresponding to this ldasGC PR are also implemented, but not yet in PR#327.  The updated version has been tested.

GEOSldas_App/ldas_setup
```diff
@@ -152,32 +155,79 @@ class LDASsetup:
...
+           self.rqdExeInp['EXP_DOMAIN']               = 'CF'+ self.adasdomain +'x6C_GLOBAL'
+           self.rqdExeInp['BCS_RESOLUTION']           = 'CF'+ self.adasdomain +'x6C_CF'     + self.adasdomain +'x6C'
+           self.rqdExeInp['RESTART_DOMAIN']           = 'CF'+ self.adasdomain +'x6C_GLOBAL' 
+           self.rsdir                                 = os.path.dirname(self.rstdir)
+           self.rqdExeInp[ 'RESTART_PATH' ]           = self.rsdir 
+           self.rqdExeInp[ 'RESTART_ID'   ]           = os.path.basename(self.rstdir)
+           # the following are not in default rqdExeInp list 
+           self.met_hinterp                           = int(self.rqdExeInp.get('MET_HINTERP',0))
+           self.rqdExeInp['MET_HINTERP']              = 0
+           self.landassim_dt                          = str(self.rqdExeInp.get('LANDASSIM_DT', "10800"))
```
gmao-rreichle, line 220: @saraqzhang, @weiyuan-jiang : Do we need `self.landassim_dt` and `self.landassim_t0` ?

GEOSldas_App/ldas_setup
```diff
@@ -152,32 +155,79 @@ class LDASsetup:
...
+           self.rqdExeInp['LANDASSIM_DT']             = "10800"
+           self.landassim_t0                          = str(self.rqdExeInp.get('LANDASSIM_T0', "013000"))
+           self.rqdExeInp['LANDASSIM_T0']             = "013000"
+           jobsgmtdef                                 = "00000000 060000"
```
gmao-rreichle, line 224: @saraqzhang , @weiyuan-jiang : Do we need `jobsgmtdef` and `self.jobsgmt` ? Can we just use
`self.rqdExeInp['JOB_SGMT'] = "00000000 060000"`

GEOSldas_App/ldas_setup
```diff
@@ -152,32 +155,79 @@ class LDASsetup:
...
+           self.rqdExeInp['BCS_RESOLUTION']           = 'CF'+ self.adasdomain +'x6C_CF'     + self.adasdomain +'x6C'
+           self.rqdExeInp['RESTART_DOMAIN']           = 'CF'+ self.adasdomain +'x6C_GLOBAL' 
+           self.rsdir                                 = os.path.dirname(self.rstdir)
+           self.rqdExeInp[ 'RESTART_PATH' ]           = self.rsdir 
+           self.rqdExeInp[ 'RESTART_ID'   ]           = os.path.basename(self.rstdir)
+           # the following are not in default rqdExeInp list 
+           self.met_hinterp                           = int(self.rqdExeInp.get('MET_HINTERP',0))
+           self.rqdExeInp['MET_HINTERP']              = 0
+           self.landassim_dt                          = str(self.rqdExeInp.get('LANDASSIM_DT', "10800"))
+           self.rqdExeInp['LANDASSIM_DT']             = "10800"
```
gmao-rreichle, line 221: @weiyuan-jiang : ldas_setup starts with a list of "required" input keys `rqdExeInpKeys` and `rqdExeInpKeys_rst`.  But later on we keep adding keys to the `rqdExeInp` list that aren't strictly "required".  For example, here we add `LANDASSIM_DT`, which isn't in the original list of "required" keys.  Maybe I'm misunderstanding the meaning of "required" here, but I would expect something more along the lines of:
1. define the list of "required" inputs
2. start a list of all rc variables, which includes the required keys and any non-required additional keys.  
Not sure if this comment makes any sense.  I'm just having a difficult time understanding the logic behind the script.

gmao-rreichle (commented):
@saraqzhang, @weiyuan-jiang:  See inline comments below.  
@weiyuan-jiang, the changes in this PR (#94) supplement what was merged in #81. Together, the two PRs modify `ldas_setup` by adding command line arguments for the benefit of setting up LDAS within the ADAS, where ldas_setup is called from within fvsetup (see https://github.com/GEOS-ESM/GEOSadas/pull/327 and https://github.com/GEOS-ESM/GEOSadas/pull/352).
@weiyuan-jiang, in addition to addressing the inline comments below, please take a look at the aforementioned PRs (#81 and #94) and make sure the modifications to `ldas_setup` are consistent with the overall design of `ldas_setup`.  Thanks!

GEOSldas_App/ldas_setup
```diff
@@ -70,59 +70,62 @@ class LDASsetup:
             # sample sub-command
             # by construction, we can have
             # either: {'exeinp': False, 'batinp': 'lasgh'}    <-- 'lasgh'???
-            # or:     {'exeinp': True, 'batinp': None}
+            # or:     {'exeinp': True,  'batinp': None   }
```
gmao-rreichle, line 73: @weiyuan-jiang : Any idea what `'lasgh'` in line 72 might mean?  Should this be replaced with True?  Also, is there a difference between "False" in line 72 and "None" in line 73?

GEOSldas_App/ldas_setup
```diff
@@ -152,32 +155,79 @@ class LDASsetup:
...
+            self.rqdExeInp['LADAS_COUPLING'] = self.lacase 
+        if cmdLineArgs['agcm_res'] != 'None' : 
+            res_string = self.agcm_res 
+            if len(res_string) == 4 : 
+               adres = res_string.replace("C", "0") 
+            if len(res_string) == 3 : 
+               adres = res_string.replace("C", "00")
+            self.adasdomain = adres 
+        if cmdLineArgs['rstdir'] != 'None' :
+            rstdir_ = cmdLineArgs['rstdir'].rstrip('/')      # remove trailing '/'
```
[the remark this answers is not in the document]
weiyuan-jiang: rstrip won't break the string. If there is no trailing char, it does nothing

GEOSldas_App/ldas_setup
```diff
@@ -1723,21 +1772,40 @@ def parseCmdLine():
...
-        help='replaces BEG_DATE time in exeinp file ',
+        help='overwrites time in BEG_DATE from exeinp file',
         type=str, default='None'
         ) 
+    p_setup.add_argument(
+        '--ladas_cpl',
+        help='specifies LDAS/ADAS coupling mode: (1) ADAS deterministic member or (2) ADAS ensemble',
+        type=str, default='None'
+        )
+    p_setup.add_argument(
```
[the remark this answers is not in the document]
weiyuan-jiang: Can you post the typical BCS used in ADAS? I guess it still used the old structure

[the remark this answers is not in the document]
gmao-rreichle: Thanks, @weiyuan-jiang.    I'm quite sure the ADAS still uses the old dir layout and naming convention (@saraqzhang should be able to confirm).  But from that we may be able to extract the BCS version (perhaps with some hard-coded translations of dir names into BCS version strings), and then we can have ldas_setup use the corresponding BCS in the new layout and naming convention.  @saraqzhang should be  able to take care of this.

GEOSldas_App/ldas_setup
```diff
@@ -152,32 +155,79 @@ class LDASsetup:
...
-           if self.ladas_coupling == 2 :
-                 self.rqdExeInp[ 'EXP_ID' ] = self.adas_expid + '_LDAS4ens'
-                 self.rqdExeInp[ 'MET_PATH' ] = self.adas_expdir +'/atmens/mem'
-
-        self.first_ens_id = int(self.rqdExeInp.get('FIRST_ENS_ID',0))
+               # ldas coupled with determistic component of adas
+               self.rqdExeInp['EXP_ID']               = self.adas_expid  + '_LDAS'
+               self.rqdExeInp['MET_PATH']             = self.adas_expdir + '/recycle/holdpredout'
+               self.rqdExeInp['ENSEMBLE_FORCING']     = 'NO'
+               self.nens = 24
```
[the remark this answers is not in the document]
saraqzhang: we can remove nens from the newly minted "default values list", and keep it in exeinp.  For atmens at setup stage we dont have nens information, it is determined after the atmens ICS are installed.

GEOSldas_App/ldas_setup
```diff
@@ -152,32 +155,79 @@ class LDASsetup:
...
+           elif self.ladas_coupling == 2 :
+               # ldas coupled with ensemble component of adas
+               self.rqdExeInp['EXP_ID']               = self.adas_expid  + '_LDAS4ens'
+               self.rqdExeInp['MET_PATH']             = self.adas_expdir + '/atmens/mem'
+               self.rqdExeInp['ENSEMBLE_FORCING']     = 'YES' 
+               self.nens = 32
```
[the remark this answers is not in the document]
saraqzhang: For atmens at setup stage we dont have nens information, it is determined after the atmens ICS are installed.

GEOSldas_App/ldas_setup
```diff
@@ -1723,21 +1772,40 @@ def parseCmdLine():
         'batinpfile',
         help='input file with arguments for SLURM',
         )
+
+    # the following command line arguments, if present, take precedence over what
```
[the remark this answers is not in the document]
weiyuan-jiang: non-positional arguments start with '-' or '--'. It is better to start with positional and then non-positional arguments

GEOSldas_App/ldas_setup
```diff
@@ -152,32 +155,79 @@ class LDASsetup:
...
+               self.nens = 32
+           else :
+               exit("Error. Unknown value of ladas_coupling.\n")
+           self.rqdExeInp['NUM_LDAS_ENSEMBLE']        = self.nens
+           self.first_ens_id                          = 1
+           self.rqdExeInp['FIRST_ENS_ID']             = self.first_ens_id 
+           self.rqdExeInp['EXP_DOMAIN']               = 'CF'+ self.adasdomain +'x6C_GLOBAL'
+           self.rqdExeInp['BCS_RESOLUTION']           = 'CF'+ self.adasdomain +'x6C_CF'     + self.adasdomain +'x6C'
+           self.rqdExeInp['RESTART_DOMAIN']           = 'CF'+ self.adasdomain +'x6C_GLOBAL' 
+           self.rsdir                                 = os.path.dirname(self.rstdir)
```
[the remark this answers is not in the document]
weiyuan-jiang: Here the restdir may or may not include the restart_id, it is confusing and needs more clarification or even change the name

GEOSldas_App/ldas_setup
```diff
@@ -152,32 +155,79 @@ class LDASsetup:
...
+           self.rqdExeInp[ 'RESTART_PATH' ]           = self.rsdir 
+           self.rqdExeInp[ 'RESTART_ID'   ]           = os.path.basename(self.rstdir)
+           # the following are not in default rqdExeInp list 
+           self.met_hinterp                           = int(self.rqdExeInp.get('MET_HINTERP',0))
```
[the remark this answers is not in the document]
weiyuan-jiang: I don't think it is needed

GEOSldas_App/ldas_setup
weiyuan-jiang: Indeed, we use the word rqdExeInpKeys to hold some optional keys. We my change it to 'ExeInpKeys'  or divide it into more specific groups of keys

GEOSldas_App/ldas_setup
weiyuan-jiang: Both are logic types and by default they are Ture.  So we should use True or False

GEOSldas_App/ldas_setup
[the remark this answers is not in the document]
gmao-rreichle: @saraqzhang, if fvsetup knows the BCs version (as in $landbcs = Icarus-NLv3), we should pass this info into ldas_setup.  Then ldas_setup can then assemble the BCs info that is compatible with the LDAS.  
For LDAS, the BCs path should be /discover/nobackup/projects/gmao/bcs_shared/fvInput/ExtData/esm/tiles/, which for now can be hardwired into the exeinp file.  (There's a similar path at NAS, if we need to make it work for NAS as well.)
Upon input into ldas_setup, the BCs version "Icarus-NLv3" from fvsetup can be translated into "NL3" or "Icarus-NLv3_new_layout".   
And ldas_setup already knows the AGCM horizontal resolution.  
If we proceed in this way, the forthcoming update of the DAS to the v12 AGCM should be trivial.  The DAS with the v12 AGCM will presumably use BCs version "v13" in the new layout, so if we simply pass $landbcs="v13" from fvsetup into ldas_setup, everything should work without further modifications.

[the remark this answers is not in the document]
saraqzhang: yes, we can use fvsetup  BCs version info  in ldas_setup and assemble the BCs setting according to the version  and resolution info. I will add this to #94 here and to adas follow-up #352

saraqzhang:
added command line option to accept land BCs version info from fvsetup.

GEOSldas_App/ldas_setup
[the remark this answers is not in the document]
saraqzhang: added the command line option to accept land BCs version info  from fvsetup and assemble ldas BCs path.

GEOSldas_App/ldas_setup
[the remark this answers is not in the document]
saraqzhang: keep nens in default values list for the convenience of using one exeinp file for both det and ens coupling cases.

GEOSldas_App/ldas_setup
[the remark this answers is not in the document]
saraqzhang: changed the name to "rstloc", ( restart location)  that contains  path/restart_id,  and os.path.dirname(rstloc)  gets RESTART_PATH

GEOSldas_App/ldas_setup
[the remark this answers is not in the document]
saraqzhang: removed.

GEOSldas_App/ldas_setup
saraqzhang: edited

saraqzhang:
fvsetup only provide BCs version info, not the whole path. In exeinp we still need to specify BCS_PATH  such as:
BCS_PATH:  /discoer/nobackup/projects/gmao/bc_shared/fvInput/ExtData/esm/tiles/v11 
 if in ladas setup, ldas_setup will use  the BCs version from fvsetup  to replace the version "v11" in the string.

GEOSldas_App/ldas_setup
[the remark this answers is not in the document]
gmao-rreichle: Edited in https://github.com/GEOS-ESM/GEOSadas/pull/352/files

GEOSldas_App/ldas_setup
gmao-rreichle: Edited -- I hope my changes are correct.

gmao-rreichle:
@saraqzhang : I made a few more changes in an attempt to simplify things.  I hope I didn't break anything.  Please review my changes: https://github.com/GEOS-ESM/GEOSldas_GridComp/compare/29f77d8..7742280?w=1 (white-space changes are hidden in this link).
Also, don't we still need to update the sample files in `GEOSldas_App/sample_config_files/LADAS`?

gmao-rreichle:
> I have an updated version of exeinp.txt, I can put it to this branch served as a draft sample file in GEOSldas_App/sample_config_files/LADAS.

Yes, please!  And use "git rm ..." to delete the old sample files.

saraqzhang:
@gmao-rreichle   I add a README file in the directory,  it turned out no need a sample file, just  the information on which parameters will be taken care of by fvsetup or by ldas_setup default.

saraqzhang:
it may no longer need  this directory in the repo,  or change the dir name ?

gmao-rreichle:
> @gmao-rreichle I add a README file in the directory, it turned out no need a sample file, just the information on which parameters will be taken care of by fvsetup or by ldas_setup default.

See https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/8a7c7fcf97fc1fac6df0e8fc48280d09e16e90f3#commitcomment-156483200

GEOSldas_App/ldas_setup
```diff
@@ -152,40 +161,77 @@ class LDASsetup:
         if cmdLineArgs['nymdb'] != 'None' and cmdLineArgs['nhmsb'] != 'None' :
             self.date  = f"{self.nymdb} {self.nhmsb}"
```
weiyuan-jiang, line 162: self.date is not used and is not necessarr to be define. Combine the two line to one

GEOSldas_App/ldas_setup
```diff
@@ -1723,21 +1772,40 @@ def parseCmdLine():
...
+        '--ladas_cpl',
+        help='specifies LDAS/ADAS coupling mode: (1) ADAS deterministic member or (2) ADAS ensemble',
+        type=str, default='None'
+        )
+    p_setup.add_argument(
+        '--agcm_res',
+        help='specifies AGCM resolution of coupled ADAS component',
+        type=str, default='None'
+        )
+    p_setup.add_argument(
```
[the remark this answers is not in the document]
gmao-rreichle: I think this has been clarified in my commits of yesterday ("rstloc = restart_path/restart_id), but I'm not 100% sure it's correct...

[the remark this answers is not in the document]
saraqzhang: it is correct.

gmao-rreichle:
@weiyuan-jiang : Many thanks for the review.  I implemented the suggested changes in two commits:
[https://github.com/GEOS-ESM/GEOSldas_GridComp/compare/8a7c7fc..f07f2e4?w=1](https://github.com/GEOS-ESM/GEOSldas_GridComp/compare/8a7c7fc..f07f2e4?w=1)
I hope I got everything right and resolved the comments for now.  Please double-check, thanks!

GEOSldas_App/ldas_setup
```diff
@@ -69,60 +69,69 @@ class LDASsetup:
...
+        assert os.path.isdir(exphome_)                                   # exphome should exist
+        self.exphome            = os.path.abspath(exphome_)
+        self.verbose            = cmdLineArgs['verbose']
+
+        # command line args for coupled land-atm DAS
+        self.nymdb              = cmdLineArgs['nymdb']
+        self.nhmsb              = cmdLineArgs['nhmsb']
+        self.agcm_res           = cmdLineArgs['agcm_res']
+        self.bcs_version        = cmdLineArgs['bcs_version']
+        self.ladas_cpl          = cmdLineArgs['ladas_cpl']
```
gmao-rreichle, line 98: @weiyuan-jiang , @saraqzhang : I'm a bit confused about the need for self.ladas_cpl and also self.ladas_coupling defined below.  I assume there's a way to use just one local variable, but I'm not sure I would get the order of operations right.  The idea is to have ladas_cpl from the command line take precedence over what may be defined in the exeinp file.   Which makes me wonder if we even need to allow for LADAS_COUPLING to be defined in the exeinp file.   It looks like more cleanup is needed here.

GEOSldas_App/ldas_setup
saraqzhang: this  inline comment is not correct: "# when coupled to ADAS, "BCS_PATH" in exeinp file must EXCLUDE bcs version info" . In exeinp file BCS_PATH remains the same way regardless coupled or not, with version as the last element of the string.  In coupled case  tmppath_ = os.path.dirname(self.rqdExeInp['BCS_PATH'].rstrip('/')) will take  version off  and later re-assemble the PATH with tmppath_ +  '/' + self.bcs_version  ( self.bcs_version from fvsetup) .

gmao-rreichle: Thanks for pointing this out, @saraqzhang.  I think it's better to require the BCS_PATH without BCS_VERSION in the exeinp file, so the user doesn't specify a version that would then be ignored/overwritten.  After adding the comment yesterday, I forgot to fix the python code, which I just attempted to do in the following commit: https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/0ec5846c7842bf8058a8a2c59f7108477a045bce
I hope this will work now.  
Having said that, we should keep in mind that BCS_PATH should really come from fvsetup as well.  The reason it doesn't for now is because fvsetup is behind the times and still uses the bcs in the old layout.  Perhaps the better approach would be to hardwire the translation of the BCS_PATH into fvsetup (as part of https://github.com/GEOS-ESM/GEOSadas/pull/352/).  (Then we might as well translate BCS_VERSION in fvsetup as well.)  Or we could translate BCS_PATH inside ldas_setup.

saraqzhang: Currently BCS_PATH  is rather different between ldas and adas.  
ldas:  /discover/nobackup/projects/gmao/bcs_shared/fvInput/Extdata/esm/tiles/Icarus-NLv3_new_layout/land/CF360x6C_CF360x6C
adas: /discover/nobackup/projects/gmao/Share/gmaoops/fvInput_4dvar/g5gcm/bcs/Icarus-NLv3/Icarus-NLv3_ostia/CF360x6C_CF360x6C 
 
 Eventually  will the BCS location  be defined by adas or   will ldas  always keep its own ?  For now I think it is still better to do the translation in ldas_setup.

GEOSldas_App/ldas_setup
saraqzhang: @gmao-rreichle    that's right , we do not need LADAS_COUPLING in exeinp file.

saraqzhang: @gmao-rreichle    in current version:  (1)  ladas_cpl gets the command arg.  (2) self.rqdExeInp['LADAS_COUPLING'] = self.ladas_cpl  (3) self.ladas_coupling = int(self.rqdExeInp.get('LADAS_COUPLING',0))  (4) self.ladas_coupling  controls a few things down the process   and (5) LADAS_COUPLING the value written into LDAS.rc.

saraqzhang: @gmao-rreichle    we probably can directly use self.ladas_coupling = cmdLineArgs('ladas_cpl')  , eliminate self.ladas_cpl

saraqzhang:
@gmao-rreichle    the commit ea03973  has the implementation according to the spreadsheet of inp variables.  For ladas the assumption is to run  ldas_setup sample  first to produce inp files, then run ldas_setup  setup with all command line inputs and the inp files just produced.

gmao-rreichle (commented):
@saraqzhang, @weiyuan-jiang : 
- I edited Sara's most recent changes a bit, see https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/7f115d9f865c24312f11712cf4e99087b6b156d8
- I also added a few inline comments below. 
- @saraqzhang: There is at least one more instance of "GEOSldas=>" here:
https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/7f115d9f865c24312f11712cf4e99087b6b156d8/GEOSldas_App/ldas_setup#L638
I think in this place we're only extracting a list of keys (as opposed to the default parameter values), but I'm not sure.  We'll need to review this in a bit more detail.

gmao-rreichle:
@saraqzhang:  Here's a draft list of things we'll need to discuss with Ricardo:
- Can ldas_setup obtain the number of ens members from fvsetup?
- How should we go about the SLURM parameter defaults (n-tasks-per-node, n-tasks)?
- Where should we keep the MWRTM parameters and NML input files? 
- Can ldas_setup obtain "DTSTEP_ADAS_CYCLE" from fvsetup? (probably trivial)
- Is it ok to write the sample exeinp file to disk and then read it back immediately?  (Perhaps we can tweak the python routines to write the sample exeinp file for the LADAS into memory, that is, write the output from "ldas_setup sample" into a py object that "ldas_setup setup" can use.  This may be a question for Weiyuan.

GEOSldas_App/ldas_setup
```diff
@@ -1118,15 +1194,15 @@ class LDASsetup:
                     print (line.rstrip().replace('GEOSldas_expid',self.rqdExeInp['EXP_ID']))
 
             # take  HISTdet or HISTens if ladas 
-            if shortfile =='HISTdet.rc' and self.ladas_coupling == 1 :
+            if shortfile =='HISTdet.rc' and self.ladas_cpl == 1 :
```
gmao-rreichle, line 1197: @saraqzhang, @weiyuan-jiang: I'm trying to understand the logic of the code here.  It looks like ldas_setup first turns the default GEOSldas_HIST.rc into HISTORY.rc, then, if there is a HISTRC_FILE specified, the HISTRC_FILE overwrites the previously generated HISTORY.rc, then, if self.ladas_cpl==1, ldas_setup does the same with GEOSldas_HISTdet.rc (thereby overwriting the previously generated HISTORY.rc), and then, if self.ladas_cpl==2, ldas_setup does the same with GEOSldas_HISTens.rc (thereby overwriting the previously generated HISTORY.rc).  My sense is that the logic of this could be improved.  Also, I'm wondering if this allows for the user to specify their own HISTORY.rc in the case of self.ladas_cpl>0.  This probably needs further investigation.

saraqzhang: the idea ( logic)  is that  for ladas case the user is not allowed to specify his own HISTORY.rc, at least at the setup process.  After the whole thing is setup, of course the user can go to the run directory and modify the HISTORY.rc installed there if he wanted , assuming he knows not to comment out the essential collection such as incr.

saraqzhang: with the current development for ladas setup, nothing is specified including HISTC_FILE in the exeinp generated by ldas_setup sample,  the result will always be HISTdet.rc or HISTdens,rc  into HISTORY.rc in run/.

saraqzhang:
it should be ok to write  sample inp to a directory ( in  current fvsetup we already inquire the directory path where the inp files are) , then read it back immediately in next call ldas_setup .

saraqzhang:
@gmao-rreichle added command line inputs:  adas_cyc,  specnml_path, mwrtm_path

saraqzhang:
@gmao-rreichle  I have not yet added the command to call ldas_setup sample  in fvsetup.  there are some implications :  in ldas_setup _printExeInputKeys() has " _fn= ../etc/GEOSldas_LDAS.rc " . For ldas_setup sample called in runjob (fvsetup), we need  a complete path, otherwise it could not find etc/ .   Should we add command line input option for the path and modify _printExeInputKeys() ?

saraqzhang:
@gmao-rreichle @weiyuan-jiang    with the merge and the fixes up to this point  the setup test failed at:  when creating links in input/ such as  tile.data -> /discover/nobackup/projects/ir-cld/qzhang/gdas/hybd_ladas/lana/hybd_ladas_LDAS/output/CF0360x6C_GLOBAL/rc_out//CF0360x6C_CF0360x6C-Pfafstetter.til.domain , it added "domain"  in the filename.  ( all links in input/ have the same wrong addition ) .

weiyuan-jiang:
> @gmao-rreichle @weiyuan-jiang with the merge and the fixes up to this point the setup test failed at: when creating links in input/ such as tile.data -> /discover/nobackup/projects/ir-cld/qzhang/gdas/hybd_ladas/lana/hybd_ladas_LDAS/output/CF0360x6C_GLOBAL/rc_out//CF0360x6C_CF0360x6C-Pfafstetter.til.domain , it added "domain" in the filename. ( all links in input/ have the same wrong addition ) .

The suffix "domain" means the simulation is not in a global domain

gmao-rreichle:
> This is the old way to check if it is zoomin or not. I may mess up the logic here
> 
> https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/5d6e6185b11a4bd1ab5d4ae52b6c7073e096c2ea/GEOSldas_App/ldas_setup#L751-L754
> 
> )

@weiyuan-jiang : Which branch are you quoting from above?  Hasn't the variable self.islocal been replace by self.isZoomIn?  In my commit of yesterday (after your conflict resolution), I changed on instance of self.islocal to self.isZoomIn.  https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/86056e5676f240fe2d22cb31dbe0e59fc10bfd4d.  Maybe my change wasn't right?

weiyuan-jiang:
> > This is the old way to check if it is zoomin or not. I may mess up the logic here
> > https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/5d6e6185b11a4bd1ab5d4ae52b6c7073e096c2ea/GEOSldas_App/ldas_setup#L751-L754
> > 
> > )
> 
> @weiyuan-jiang : Which branch are you quoting from above? Hasn't the variable self.islocal been replace by self.isZoomIn? In my commit of yesterday (after your conflict resolution), I changed on instance of self.islocal to self.isZoomIn. [86056e5](https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/86056e5676f240fe2d22cb31dbe0e59fc10bfd4d). Maybe my change wasn't right?

My quote is from 'main' branch. But the develop branch has already been using the other way to check the f2g file and it is fine. I don't have Sara's exeinp file and cannot see the problem

saraqzhang:
@weiyuan-jiang  what is the variable  in exeinp file  for this choice ?  in our new approach for coupled case ( PR #94) , exeinp   contains only the variables copied from geosldas.rc and surface.rc,   not rest of the variables originally in the first block  generated in "sample" .

GEOSldas_App/ldas_setup
```diff
@@ -65,77 +66,93 @@ class LDASsetup:
...
-        self.bcs_landshared = ''
-        self.tile_types = ''
-        self.with_land = False
-        self.with_landice = False
+        self.has_geos_pert      = False
+        self.nSegments          = 1
+        self.perturb            = 0
+        self.first_ens_id       = 0
+        self.in_rstfile         = None
+        self.in_tilefile        = 'None'                 # default string
```
gmao-rreichle, line 140: @weiyuan-jiang , @saraqzhang : Why do we use None (without quotation marks) for self.in_rstfile and 'None' (with quotation marks) for self.in_tilefile?  Can't we make this consistent?

GEOSldas_App/ldas_setup
```diff
@@ -504,21 +579,28 @@ class LDASsetup:
...
-        assert self.rqdRmInp['walltime']
-        ## ntasks_model is a +ve integer
-        _ntasks = int(self.rqdRmInp['ntasks_model'])
-        assert _ntasks>0
-        self.rqdRmInp['ntasks_model'] = _ntasks
-        _ntasks = None
+        if self.ladas_cpl > 0 : 
+           if cmdLineArgs['account'] != 'None':
+                self.rqdRmInp['account'] = cmdLineArgs['account']
+           self.rqdRmInp['walltime']        = "00:30:00"
```
gmao-rreichle, line 585: @saraqzhang : Are we sure that 30 min is sufficient for c360 and, eventually, c720 or higher?  Maybe increase to 1 hour?

GEOSldas_App/ldas_setup
```diff
@@ -151,15 +168,94 @@ class LDASsetup:
...
+            _end_date = _beg_date + timedelta(hours=self.varwindow/60) 
+
+            self.rqdExeInp['END_DATE']                 = _end_date.strftime("%Y%m%d %H%M%S") 
+            #default or input from command line 
+            self.rqdExeInp['NML_INPUT_PATH']           = "/discover/nobackup/qzhang/LDAS_INPUTS/SPECnml/" 
+            if self.LDAS_nml_path != 'None' :
+                self.rqdExeInp['NML_INPUT_PATH']       = self.LDAS_nml_path
+            self.rqdExeInp['MWRTM_PATH']               = "/discover/nobackup/qzhang/LDAS_INPUTS/MWRTM/" 
+            if self.mwrtm_path != 'None' :
+                self.rqdExeInp['MWRTM_PATH']           = self.mwrtm_path
```
gmao-rreichle, line 250: I think we should drop the default values here.  If we hardwire these paths anywhere, it should be in fvsetup.

gmao-rreichle (commented):
@saraqzhang: I made a few more changes (https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/ae2717e9bddc2ea11ba9fcc96da929b8abfdb23a). See also comments/questions below.

GEOSldas_App/ldas_setup
weiyuan-jiang: 'None' is a str. If you don't provide default in the command line argument, it will be None ( a type in python). For example, here https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/ae2717e9bddc2ea11ba9fcc96da929b8abfdb23a/GEOSldas_App/ldas_setup#L1915
 The test would be like
if cmdLineArgs['account'] is not None:

saraqzhang:
@gmao-rreichle   additional edits commit 8dbd52a.   Note that the test on the current version fails due to  isZoomin  issue as  seen yesterday.

GEOSldas_App/ldas_setup
saraqzhang: dropped default values in ldas_setup.  default values in fvsetup

gmao-rreichle:
> @gmao-rreichle additional edits commit [8dbd52a](https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/8dbd52adec16e0f55dce699fa9925b2eae948c78). Note that the test on the current version fails due to isZoomin issue as seen yesterday.

Thanks, @saraqzhang.  Did you recreate the sample exeinp file before running the test?  I'm wondering if a change associated with the merging of the landice PR might have changed the sample exeinp file.  (We are still missing the call to "ldas_setup sample" in GEOSadas PR#352.)

Also, I tried to clean up the processing of the LDAS or AGCM defaults from GEOS_SurfaceGridComp.rc: https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/29a2b3cc23e93eb4c183cad831f6b882ede675d6

gmao-rreichle:
> @gmao-rreichle I have not yet added the command to call ldas_setup sample in fvsetup. there are some implications : in ldas_setup _printExeInputKeys() has " _fn= ../etc/GEOSldas_LDAS.rc " . For ldas_setup sample called in runjob (fvsetup), we need a complete path, otherwise it could not find etc/ . Should we add command line input option for the path and modify _printExeInputKeys() ?

@saraqzhang, @weiyuan-jiang : I think it may be better to try and modify ldas_setup so that it will produce a dictionary when called with "ladas_cpl>0".   That is, when ldas_setup is called with "ladas_cpl>0", we don't require an exeinp file and obtain the necessary information directly form the default rc files.  This isn't trivial, but it's much better than writing and subsequently reading a custom sample exeinp file.

weiyuan-jiang:
> > @gmao-rreichle I have not yet added the command to call ldas_setup sample in fvsetup. there are some implications : in ldas_setup _printExeInputKeys() has " _fn= ../etc/GEOSldas_LDAS.rc " . For ldas_setup sample called in runjob (fvsetup), we need a complete path, otherwise it could not find etc/ . Should we add command line input option for the path and modify _printExeInputKeys() ?
> 
> @saraqzhang, @weiyuan-jiang : I think it may be better to try and modify ldas_setup so that it will produce a dictionary when called with "ladas_cpl>0". That is, when ldas_setup is called with "ladas_cpl>0", we don't require an exeinp file and obtain the necessary information directly form the default rc files. This isn't trivial, but it's much better than writing and subsequently reading a custom sample exeinp file.

Sara, you done't need to pass in extra path. suggest change:

    current_directory = os.path.dirname(__file__)
    _fn = current_directory[0:-6]+'/etc/GEOSldas_LDAS.rc'

saraqzhang:
@gmao-rreichle   @weiyuan-jiang  the current status of tests:  
(1) "offline ldas_setup setup"  -- not called from fvsetup, but  run from bin/  with all command line inputs constructed like we have in fvstup--   failed  because of isZoomin issue after the landice merge. The exeinp used was  from ldas_setup sample generated last week before merge and other edits. 
(2) "offline ldas_setup sample --ladas_cpl 1 "  -- not called from fvsetup, but  run from bin/ --   there is problem in the sample exeinp file generated in current version. it keeps the line such as " #GEOSldas=> SURFLAY: 50 "  in the file  even though  GEOSagm => line is chosen and written as SURFLAY: 50 in the file. Later on ldas_setup setup failed to parse the exeinp file with these extra #lines . 
(3) "integrated " setup  -- e.g. runjob , ldas_setup called from fvsetup with command line inputs --  I will try the modification on the path that Weiyuan suggested here.

weiyuan-jiang:
> 

Sara, I can help to generate the dictionary internally. Basically, the setup would look like
ldas_setup setup path_to_simulation  NOEXEINPUT_file  slurm_file  --positional arguments

saraqzhang:
@weiyuan-jiang  that would be great. Does it mean ldas_setup setup can skip  exeinp file and get the information from those rc files directly ?   As I  mentioned above we have problem in the current version of writing exeinp with extra lines dealing with variables from surface.rc .

gmao-rreichle:
>  As I mentioned above we have problem in the current version of writing exeinp with extra lines dealing with variables from surface.rc .

I'm getting a bit lost on the details here, but if the presence of 'GEOSldas=>' is the issue, maybe try again after my most recent change   https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/29a2b3cc23e93eb4c183cad831f6b882ede675d6  ?

saraqzhang:
@gmao-rreichle   the extra lines occurred in the version with commit  29a2b3c in.  (  I  only edited  the indentation of line 692  linenum += 1  to run the test )

gmao-rreichle:
>  I only edited the indentation of line 692 linenum += 1 to run the test 

Oops, sorry!  I just fixed the indent (https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/1352dbb5d4a9c0a687e4ac798fa5be741b79d022)

Re. the presence of '# GEOSldas=>', we probably need to rethink the parsing of the SurfaceGridComp.rc file.

gmao-rreichle:
> @gmao-rreichle I made modification to skip the extra lines in writing to sample exeinp. I am still testing the path to etc/*rc in running sample from fvsetup according to Weiyuan's suggestion.

@saraqzhang : I have to think a bit more about this commit, but I have a feeling it can still be simplified further. Also, Weiyuan and I successfully got ldas_setup to work this morning (after fixing the py errors I had introduced yesterday).  That is, we didn't have the "domain" problem.  I wonder if perhaps you were working with an older build?  The "landice" PR changes require a fresh build.  Also, Weiyuan and I discussed a strategy for making "ldas_setup [...] --ladas_coupling [1,2]" work without needing a sample exeinp file.  Weiyuan is going to take a crack at implementing it.

saraqzhang:
@gmao-rreichle  sounds great if we can skip exeinp in coupled case.  I will wait for the updated version. will it be pushed to this branch ( with PR #94) , or a new one?

weiyuan-jiang:
> @gmao-rreichle sounds great if we can skip exeinp in coupled case. I will wait for the updated version. will it be pushed to this branch ( with PR #94) , or a new one?
I will push directly to this branch

weiyuan-jiang:
@saraqzhang I have pushed my changes but have not tested it.  Now exeinp can be a dummy string during setup when ladas_cppl > 0 ( The dummy string must be there since it is positional argument). The self.rqdExeInp dictionary would be produced without a real exeinp file. Hopefully it would work. I can debug with you if anything goes wrong.

saraqzhang:
@weiyuan-jiang Weiyuan, thanks for the update, I will make tests today.

saraqzhang:
@weiyuan-jiang @gmao-rreichle  I made minor changes  ( df1b1f3 ) and  ran setup test with  dummyexeinp   and batinp.txt  ( generated by previous sample run) in position arguments,  the test successfully setup the exp.

gmao-rreichle:
@saraqzhang, @weiyuan-jiang : I made more changes (https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/4a1802eedef0a16a708a6554a41204596c6dadfa).  Besides editing comments, I tried to clean up the logic of how resource manager inputs are processed.  I also (tentatively) reverted back to relative paths for  the location of the rc template files.  With the absolute paths, the sample --exeinp command broke for me, and I think that ldas_setup should in any case be run from the ./install dir, so ../etc/ should always point to the rc template files.  I have done some limited testing, but I can't promise that everything will work.  We may have to iterate more.

weiyuan-jiang:
> > @saraqzhang, @weiyuan-jiang : I made more changes ([4a1802e](https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/4a1802eedef0a16a708a6554a41204596c6dadfa)). Besides editing comments, I tried to clean up the logic of how resource manager inputs are processed. I also (tentatively) reverted back to relative paths for the location of the rc template files. With the absolute paths, the sample --exeinp command broke for me, and I think that ldas_setup should in any case be run from the ./install dir, so ../etc/ should always point to the rc template files. I have done some limited testing, but I can't promise that everything will work. We may have to iterate more.
> 
> @gmao-rreichle Can you post the broken message? Or simply print out current_directory = os.path.dirname(**file**). There are two underline before and after file. it does not show here. Sara needs absolute path.  Otherwise it is relative to fvsetup directory

gmao-rreichle:
> @gmao-rreichle Can you post the broken message? Or simply print out current_directory = os.path.dirname(**file**). There are two underline before and after file. it does not show here. Sara needs absolute path. Otherwise it is relative to fvsetup directory

@weiyuan-jiang, @saraqzhang : I realized that I was running ldas_setup from a directory that was not called "./bin" but had a longer name. (I just copied the key executables from a build by Matt's system tests.)  In my testing, I then keep the rc template files in a directory ../etc/.   The "full path" approach assumes that the current dir has a name that is 3 chars long `current_directory[0:-4]`, and that the rc template files are in ../etc relative to the current path.   After renaming my current directory to "bin", the "full path" approach works for me.  But I don't see how this "full path approach" is different from just specifying the relative path. @saraqzhang : what is your error message when it fails with the relative path approach.

weiyuan-jiang:
> > @gmao-rreichle Can you post the broken message? Or simply print out current_directory = os.path.dirname(**file**). There are two underline before and after file. it does not show here. Sara needs absolute path. Otherwise it is relative to fvsetup directory
> 
> @weiyuan-jiang, @saraqzhang : I realized that I was running ldas_setup from a directory that was not called "./bin" but had a longer name. (I just copied the key executables from a build by Matt's system tests.) In my testing, I then keep the rc template files in a directory ../etc/. The "full path" approach assumes that the current dir has a name that is 3 chars long `current_directory[0:-4]`, and that the rc template files are in ../etc relative to the current path. After renaming my current directory to "bin", the "full path" approach works for me. But I don't see how this "full path approach" is different from just specifying the relative path. @saraqzhang : what is your error message when it fails with the relative path approach.

I believe the current directory is where fvsetup sits when Sara runs fvsetup. So the relative path would not work.

gmao-rreichle:
> I believe the current directory is where fvsetup sits when Sara runs fvsetup. So the relative path would not work.

So fvsetup isn't run from the install directory?  And if it isn't, why do we expect the rc files in [abspath]/etc/?  Sara, please point me to a build of the DAS (with the integrated LDAS).

weiyuan-jiang:
> > I believe the current directory is where fvsetup sits when Sara runs fvsetup. So the relative path would not work.
> 
> So fvsetup isn't run from the install directory? And if it isn't, why do we expect the rc files in [abspath]/etc/? Sara, please point me to a build of the DAS (with the integrated LDAS).

The absolute path is ldas_setup's path.

gmao-rreichle:
I just pushed another commit https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/c3ca2d5f08366ba20b09a7c1dda486b17fd7f805
Do you think this will work?  I was trying to explain in comments what we're doing ,and it was difficult to explain [0:-4] without a lot of words

weiyuan-jiang:
> I just pushed another commit [c3ca2d5](https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/c3ca2d5f08366ba20b09a7c1dda486b17fd7f805) Do you think this will work? I was trying to explain in comments what we're doing ,and it was difficult to explain [0:-4] without a lot of words

I think it should work

gmao-rreichle:
@weiyuan-jiang, @saraqzhang : I added one more commit (https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/94/commits/0f220cb7d946d85b9d377b5cfccd07e906463ada) in an attempt to clean up the logic of processing the default values from the GEOS_SurfaceGridComp.rc file.  I did some limited testing but can't promise for sure it's flawless.  When you get a chance, please test and/or take a look at the mods.  Thanks!

weiyuan-jiang:
> @weiyuan-jiang, @saraqzhang : I added one more commit ([0f220cb](https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/0f220cb7d946d85b9d377b5cfccd07e906463ada)) in an attempt to clean up the logic of processing the default values from the GEOS_SurfaceGridComp.rc file. I did some limited testing but can't promise for sure it's flawless. When you get a chance, please test and/or take a look at the mods. Thanks!
> 
> PS: I just remembered that we have essentially the same logic in two places:
> 
> 1. In function `_parseInputFile(...)`: https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/3d9121e9d80e8bf4efa6fa254bb367fc3663d55f/GEOSldas_App/ldas_setup#L678
> 2. In function `_ProduceExeInput(...)`: https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/3d9121e9d80e8bf4efa6fa254bb367fc3663d55f/GEOSldas_App/ldas_setup#L1803
> 
> I suppose we can clean this up by creating a function for processing GEOS_SurfaceGridComp.rc. Maybe we can leave this for the bigger cleanup of ldas_setup.

Yes. I will clean that up

PR #96, opened 2025-05-19
Add model-based QC for MODIS SCF observations

amfox37:
Analysis of MODIS Aqua and Terra snow cover fraction (SCF) observations showed many "bad" (high) observations were passing existing QC using observation QC flags, even when using the most conservative settings. 
Following existing templates for model-based QC we have added a subroutine to replace predicted `asnow` observations with a  missing value when temperature of soil layer 1 exceeds a hardwired threshold. 
After some sensitivity testing this threshold is currently set to 10 degrees C. 

Successfully 0-diff tested by @gmao-rreichle after https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/96/commits/59aa257eb4eeb226ae3239e82ab9e5aa8534c21b
(For tests with aggressive builds, the "catparam" file comparison failed within roundoff for **cdcr1**.)


gmao-rreichle:
Thanks, @amfox37, for putting this together.  I just added a small change (https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/96/commits/59aa257eb4eeb226ae3239e82ab9e5aa8534c21b) which I think is critical when running **only** SCF assimilation.   Please double-check. 
cc: @lcandre2

amfox37:
> Thanks, @amfox37, for putting this together. I just added a small change ([59aa257](https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/59aa257eb4eeb226ae3239e82ab9e5aa8534c21b)) which I think is critical when running **only** SCF assimilation. Please double-check. cc: @lcandre2

Yes, I've double checked and this is now working when you only have SCF obs and with update_type 11. My bad for not checking this before - obviously I had multiple obs types in the obs_param namelist so`get_tp_l` was being called...

gmao-rreichle (approved):

PR #101, opened 2025-06-02
fixed UDUNITS error

weiyuan-jiang:
Without this PR, GEOSldas run got an error
pe=00002 FAIL at line=00047    ApplicationSupport.F90                   <status=102>
pe=00004 FAIL at line=00114    MAPL_Cap.F90 

It will not affect the final results

### DIFF
--- a/GEOSldas_App/lenkf_j_template.py
+++ b/GEOSldas_App/lenkf_j_template.py
@@ -63,6 +63,8 @@ setenv argv
 
 source $GEOSBIN/g5_modules
 
+setenv BASEBIN ${{BASEDIR}}/Linux/bin
+
 setenv MPI_STACK {DETECTED_MPI_STACK}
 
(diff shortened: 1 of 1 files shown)


mathomp4:
@weiyuan-jiang I mean...maybe? I don't know if the nco modulefile adds more things that just to PATH?

Note, nco is in baselibs so you could try that. You'd just need to add the Baselibs bin path to path. 

Your change was the simplest though with less chance of side effects.

weiyuan-jiang:
> @weiyuan-jiang I mean...maybe? I don't know if the nco modulefile adds more things that just to PATH?
> 
> Note, nco is in baselibs so you could try that. You'd just need to add the Baselibs bin path to path.
> 
> Your change was the simplest though with less chance of side effects.

Great. I can get the command from baselib.  Now I remove the loading of nco

PR #104, opened 2025-06-05
Update some defaults in LDASsa_DEFAULT_inputs_*.nml files

gmao-rreichle:
Some of the default values for obs parameters and forcing perturbations in the LDASsa_DEFAULT_inputs_*.nml were out of sync with the values used in the latest SMAP L4_SM Version 8 algorithm.  This PR updates the default values to match those in L4_SM.


gmao-rreichle:
@gmao-qliu, @amfox37 : I started this PR in an attempt to clean up the default values of the obs parameters.  Ideally, users only need to turn on/off the "getinnov", "scale", and "assim" flags and don't have to worry about the nitty-gritty settings.  
@amfox37 : Please take a look at the SMAP Tb species parameters and compare to what you use for SMOS assimilation.  We probably need to update the SMOS parameters accordingly. 
@gmao-qliu : For SMOS ("fit"), what is the correct RTM_id?  Should we use RTM_id=2 or 3?  

```
! %RTM_ID         = ID of radiative transfer model to use for Tb forward modeling
!                   (subroutine get_obs_pred()) 
!                     0 = none
!                     1 = L-band tau-omega model as in De Lannoy et al. 2013 (doi:10.1175/JHM-D-12-092.1) (SMOS)
!                     2 = same as 1 but without Pellarin atm corr (SMAP)
!                     3 = same as 1 but with Mironov and SMAP L2_SM pol mixing (SMOS)
!                     4 = same as 3 but without Pellarin atm corr (targeted for SMAP L4_SM Version 8)

```

amfox37:
For the SMOS ("fit") values I used the values that I _think_ @gmao-qliu suggested having used them in her previous testing 6 months ago, which are very like the SMAP values. I don't recall making any changes myself, and I must admit I hadn't  looked at the previous default values until just now. If we should be using RTM_id 2 or 3 then I've been using the wrong one as I've used 4. Also, the errstd I've used is 4 and more than double than the previous default (1.5)... 

If these updated values I have just used are the settings that we actually want, the changes required to the default look like:

`%RTM_ID         = 2` changes to `= 4`
`%coarsen_pert   = .false.` changes to `=.true.`
`%errstd         = 1.5` changes to `=4.0`
`%path           = '/discover/nobackup/projects/gmao/ssd/land/l_data/SMOS/EASEv2/ESA_REPR/SMOS_M36_SCLF1C_fit_nosky_noatm_v620_ESA_v102/SMOS_fit_poly2/'` changes to `= '/discover/nobackup/projects/gmao/smap/SMAP_Nature/SMOS/EASEv2/ESA/SMOS_M36_SCLF1C_fit_nosky_noatm_v724    _ESA_v102/SMOS_fit_poly2/'`

gmao-qliu:
I think it makes sense to use the same RTM_ID setting for SMOS as for SMAP, since we only use one set of mwRTM.

amfox37:
I can confirm that the default setting here match what was used in the land sweeper runs for MODIS scf and ASCAT sfds - with the minor exception that I was using incorrect orbit flags for ASCAT. It should be 3 (asc and desc) as per the default. But I believe these aren't actually used at any stage in the reading/scaling/assimilation of the ASCAT obs so won't have impacted anything. 
The CYGNSS values agree with what I've been using in testing at this stage.

gmao-qliu:
I can also confirm that the updated DEFAULT input files are good.

biljanaorescanin (approved):

PR #106, opened 2025-06-06
Specify only ntasks_model for SLURM resource request

weiyuan-jiang:
Revise ldas_setup and lenkf_j_template.py so that only "ntasks" is specified in SLURM directives.  Previously, "nodes" and "ntasks-per-node" were specified.  

By specifying only "ntasks", SLURM can efficiently allocate the appropriate number of either Milan or Cascade Lake nodes.  

Previously, by specifying "ntasks-per-node" we had to either (i) effectively constrain to Milan nodes (nodes=1, ntasks-per-node=126) or (ii), by choosing ntasks-per-node=46 and nodes=3, end up with a potentially very wasteful allocation of several Milan nodes if Cascade Lakes are too busy (depending on how exactly SLURM translates these directives into allocations).  

The downside of specifying only "ntasks" is that we can no longer set up OSERVER via ldas_setup.  But OSERVER turned out not to provide any efficiency gains for LDAS and was never used in practice.  Note that manual editing of the lenkf.j job file (after running ldas_setup) still allows for the use of OSERVER.

---

Successfully 0-diff tested by @gmao-rreichle.   For the assim tests, results were not 0-diff because of layout changes caused by the present PR.  (Results for assimilation runs have been layout-dependent for a long time.)  The non-0-diff results are within roundoff and scientifically correct. 

Impact on automated nightly tests: The line that specifies "ntasks-per-node" needs to be removed from the ./run/*.slurm batch input config files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Specify ntasks_model only for slurm resource request
+- Specify only ntasks_model for SLURM resource request.
 - Revisions for handling of Nens and special nml and mwtrm path/files in coupled land-atm DAS.
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


weiyuan-jiang:
I think we should update the night test batinp file and remove 'ntasks-per-node' item.  It is not like "daysperjob"which is almost invisible.  The unused item in the batinp file would confuse users.

PR #102, opened 2025-06-03
revisions for handling of Nens and special nml and mwtrm path/files in coupled land-atm DAS 

saraqzhang:
Revised handling of configuration parameters in coupled land-atm DAS:
- add custom special nml files for coupled land-atm DAS to GEOSldas_GridComp source code
- hardwire new mwrtm_param path (on project disk) in ldas_setup
- obtain Nens from fvsetup via command line

0-diff tests by @gmao-rreichle after commit https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/102/commits/af22d9ac25f3a9c6331799d4f349582ce34ee813: 

- Intel tests passed
- GNU tests failed comparison (but so did Matt's automated nightly tests for develop, so not a concern for this PR)

PR #107, opened 2025-06-09
refactoring ldas_setup

### LINKED ISSUE #113: GEOSldas should add ntasks-per-node as an optional input
In #106, the use of `ntasks-per-node` as a required argument was removed. And for most runs this is good, as it makes things more general.

But, @biljanaorescanin found a need to return it as an *optional* argument. She was running the LDAS at c2160. In an older job, she was using like 1260 cores and 80 cores-per-node because the job needed more memory per core.

However, currently, there is not way in the input file to specify this, so a user must do it *after* the lenkf.j file is made. If you don't, the job is killed by NCCS for using too much memory on a node.

Looking at ldas_setup, there seem to be both required and optional arguments, and I think `ntasks-per-node` is a viable optional argument.

cc @weiyuan-jiang @gmao-rreichle

weiyuan-jiang:
Initial check in. The class will be broken up into smaller steps.

This PR addresses issues #113 and #134.

**Note**: The ldas_setup changes impact #54 (irrigation), #131 (routing), and #132 (CatchCN51).  Chances are that for each of these PRs, updating ldas_setup after the present PR has been merged will result in conflicts and require manual updates.


gmao-rreichle:
@weiyuan-jiang : Many thanks for getting a start on this.  Perhaps hold off on editing ldas_setup for refactoring until #102 and #106 have been merged.  With any luck, we can merge these PRs early this week.

gmao-rreichle:
@weiyuan-jiang : When you get back to this PR, please include an **optional** SLURM/PBS parameter `ntasks-per-node`, see issue #113.  Note that unless the **user** specifies a value for  `ntasks-per-node`, the default behavior should be to **not** specify an integer value for `ntasks-per-node` in the lenkf.j job file.  
I don't know if it's possible to specify `ntasks-per-node` in lenkf.j with a blank value.  If this works, I suppose we can add `ntasks-per-node` just like `constraint`.  If a blank value for `ntasks-per-node` doesn't work in lenkf.j, we need a bit more scripting than what we do for `constraint`.

cc: @mathomp4

weiyuan-jiang:
This PR addresses issue #134 and #113

weiyuan-jiang:
@gmao-rreichle I can merge PR #132 to this PR. Should I do that ?

weiyuan-jiang:
@saraqzhang Would you please test this PR in your GEOSadas run?  I expect zero-diff. Thanks

GEOSldas_App/setup_utils.py
```diff
@@ -0,0 +1,309 @@
...
+     return {}
+        
+   inpdict = OrderedDict()
+   errstr  = "line [%d] of [%s] is not in the form 'key: value'"
+   # determine which default values to pick from GEOS_SurfaceGridComp.rc
+   # must be in the format '# GEOSldas=>' or # GEOSagcm=>
+   if ladas_cpl == 0 :
+      use_rc_defaults = '# GEOSldas=>'    # use defaults for LDAS
+   else :
+      use_rc_defaults = '# GEOSagcm=>'    # use defaults for AGCM
```
gmao-rreichle: I think using the '#' character and space as part of the string is risky.  Future edits of GEOS_SurfaceGridcomp.rc may inadvertently drop the blank character after the '#' character, which would break the present script.  Rather than starting the line with '# GEOS...', why not identify the lines that contain 'GEOS....=>' as we did before?

GEOSldas_App/setup_utils.py
```diff
@@ -0,0 +1,309 @@
...
+   return inpdict
+
+def echoInputFile(inpfile, ladas_cpl = 0):
+   """
+    Echo inpfile, ignore line starts with "## "
+   """
+   if ladas_cpl == 0 :
+      use_rc_defaults = '# GEOSldas=>'    # use defaults for LDAS
+   else :
+      use_rc_defaults = '# GEOSagcm=>'    # use defaults for AGCM
```
gmao-rreichle, line 64: see comment above

gmao-rreichle: see comment above

GEOSldas_App/setup_utils.py
```diff
@@ -0,0 +1,309 @@
...
+
+   with open (inpfile) as fin :
+     for line in fin:
+        if line.startswith("## "):
```
gmao-rreichle: Similar to the comment above, it's risky to identify the line using the '## ' string.  Perhaps we simply keep these lines in the output file?  They don't hurt there.

saraqzhang:
@weiyuan-jiang @gmao-rreichle  I tested  ldas_setup  with ladas options.  It needs one fix in setup_utils.py : add  "from datetime  import timedelta ".   After the fix the experiments  are set up correctly.

gmao-rreichle:
> @weiyuan-jiang @gmao-rreichle I tested ldas_setup with ladas options. It needs one fix in setup_utils.py : add "from datetime import timedelta ". After the fix the experiments are set up correctly.

Thanks, @saraqzhang, for taking look so quickly.  @weiyuan-jiang, I added the import statement.

GEOSldas_App/setup_utils.py
```diff
@@ -0,0 +1,312 @@
...
+   """
+   if ladas_cpl == 0 :
+      use_rc_defaults = '# GEOSldas=>'    # use defaults for LDAS
+   else :
+      use_rc_defaults = '# GEOSagcm=>'    # use defaults for AGCM
+
+   with open (inpfile) as fin :
+     for line in fin:
+        if line.startswith(use_rc_defaults):
+           line = line.split(use_rc_defaults)[1]
```
gmao-rreichle: Why can't we replace this with the following and avoid the "startswith"?

```suggestion
       if use_rc_defaults in line:
          line = line.split(use_rc_defaults)[1]
```

gmao-rreichle: Combined with `use_rc_defaults = 'GEOSldas=>'`,  this would let us avoid having the '# GEOS[xxxx]=>' format with the all-important single whitespace character being in just the right place.

weiyuan-jiang: if so, we need special codes to deal with  this line https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/a6a6f9b66c452e1afb5d1c32c12e5767a2c468b8/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/GEOS_SurfaceGridComp.rc#L10

weiyuan-jiang: Are we waiting for this PR https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1134 ?

gmao-rreichle: @weiyuan-jiang : Yes, except the relevant PR is https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1133. Scott said he will merge https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1133 (and https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1134) as soon as the CI test completes.  But there seems to be an issue with CI that cannot be related to the PRs.  I assume it will get sorted soon.  Let's wait for now.

gmao-rreichle (approved):

Issue #103, opened 2025-06-04
Stretched Grid offline ldas runs


biljanaorescanin:
During stretched grid runs, an issue arises in the history processing because `GEOSldas` does not recognize or account for stretched-grid properties during the `HISTORY.rc` handling. Restarts were not affected. 

LDAS.rc needs these 3:

`GEOSldas.STRETCH_FACTOR:   `       
`GEOSldas.TARGET_LON:    `           
`GEOSldas.TARGET_LAT:      `         

NOTE: For GCM these were added to `AGCM.rc`

PR #109, opened 2025-06-11
added stretch grid items to LDAS.rc

weiyuan-jiang:
Addresses #103


gmao-rreichle:
@biljanaorescanin : Can you please verify that the new functionality works for you?  I suppose you don't have to run the full set of tests.  I should be enough to run ldas_setup for a regular and a stretched cube-sphere tile space and verify the resulting LDAS.rc.  Thanks!

biljanaorescanin:
This PR will do what it needs to and from setup for offline stretched grid run now I have all I need to run in `LDAS.rc`.
Regression tests are zero diff, I only run Intel but I think that is enough.

biljanaorescanin (approved):

PR #111, opened 2025-06-12
minor fixes for ObsFcstAna postprocessing package

gmao-rreichle:


amfox37:
I've been using this package quite extensively over the last few days and it seems great for my use cases at least.
The only question I have is whether the Plot_stats_timeseries.py should combine all species into a single mean by default 
i.e. `OmFm  = np.nansum(OmFm*Nobsm)/Nobsm`
As this then means the spatial_stats file is a little inconsistent with the temporal_stats file which has output by species.

GEOSldas_App/util/postproc/ObsFcstAna_stats/Plot_stats_maps.py
```diff
@@ -171,6 +138,8 @@ def plot_OmF_maps(postproc_obj, stats, fig_path='./'):
 # -----------------------------------------------------------------------------------------------
     
 if __name__ == '__main__':
+
+    # Plot maps of long-term temporal stats
```
gmao-rreichle, line 142: I deleted the application of the Nmin threshold from compute_temporal_stats_from_sums().  The Nmin=20 value was far too hidden in the depths of the calculation there.  The threshold still needs to be applied here, but I didn't implement it.  First, we need to find a good place for the specification of the Nmin value.   It could be put into user_config.py, but since Nmin can be applied at the very end (when stats are plotted), we could also put it into the main() function in Plot_stats_maps.py.   More generally, we need to settle on where exactly we want the user to edit the scripts' parameters.

GEOSldas_App/util/postproc/ObsFcstAna_stats/Get_ObsFcstAna_stats.py
```diff
@@ -69,7 +69,7 @@ def main():
     # The plotting scripts can also run standalone using the individual Plot_stats_*.py scripts,
     #   as long as the monthly sum files are available.
     
-    plot_maps       = False
+    plot_maps       = False
```
gmao-rreichle, line 72: I think we need to decide if we want one main script (get monthly sums, plot maps, plot time series) or three main scripts.  As things are now, we're duplicating the plotting of maps and time series, which only confuses things.  I think it would be cleaner to remove the plot_maps and plot_timeseries from this script altogether, and to rename it "Get_ObsFcstAna_sums.py", and then have the Plot*.py files as stand-alone scripts.   (IMPORTANT: If you implement this suggestion, use "**git** mv Get_ObsFcstAna_stats.py Get_ObsFcstAna_sums.py" to retain the history of changes!)

GEOSldas_App/util/shared/python/tile_to_latlongrid.py
```diff
@@ -0,0 +1,64 @@
...
+
+    lat_grid = np.arange( -90+resolution/2,  90+resolution/2, resolution)
+    lon_grid = np.arange(-180+resolution/2, 180+resolution/2, resolution)
+
+    lon_2d, lat_2d = np.meshgrid(lon_grid, lat_grid)
+
+    grid_data = np.zeros((len(lat_grid), len(lon_grid), nf))
+    N_tilecnt = np.zeros((len(lat_grid), len(lon_grid), nf))
+
+    for v in range(nf):
```
gmao-rreichle, line 50: It would be more efficient if we do not repeatedly caculate the i,j index values in each iteration of the loop through v.  The i,j values do not depend on the field v.  I suggest making the loop through v the inner-most loop.

PR #118, opened 2025-06-23
Revised processing of HISTORY template (GEOSldas_HIST.rc)

weiyuan-jiang:
Revised and simplified processing of HISTORY template (GEOSldas_HIST.rc).
The revised HISTORY file provides examples for additional 2d HISTORY output.
When running a simulation in EASE grid tile space, 2d latlon-gridded outputs is now possible.  
When running a simulation in cubed-sphere tile space, 2d EASE grid output is now possible.

Successfully 0-diff tested by @gmao-rreichle  after https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/118/commits/1ed302925e9cd3df940c8f68d3057bdba63f9f33

New functionality (2d output from EASE tile space and 2d output on EASE grid) verified by @gmao-rreichle

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added EASE to Latlon and Cubed-Sphere to EASE history outputs
 - Added optional SLURM "constraint".
 - Added functionality to run on tile space of stretched cube-sphere grids.
 - Added python package for post-processing ObsFcstAna output.
@@ -19,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Revised processing of HISTORY template (GEOSldas_HIST.rc).
 - Switch to using EASE grid tools in MAPL.
 - Specify only ntasks_model for SLURM resource request.
 - Revisions for handling of Nens and special nml and mwtrm path/files in coupled land-atm DAS.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added EASE to Latlon and Cubed-Sphere to EASE history outputs
 - Added optional SLURM "constraint".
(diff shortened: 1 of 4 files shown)


weiyuan-jiang:
Verification of the outputs is needed

gmao-rreichle:
@weiyuan-jiang : I tried to simplify and clean up the processing of the HIST rc template in this commit https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/118/commits/65abe5ea2c0d102daccc3cc28f27c1eab8484e68

Unfortunately, this required touching `ldas_setup` again, although it's probably not too difficult to port the changes into #107 once the present PR has been finalized and merged.  Sorry about that!

I did not 0-diff test the PR, nor did I test its implied new functionality (i.e., output EASE simulations on a 2d lat/lon grid).  

Because I eliminated the "ASSIM" flag of process_hist.csh, it's possible my changes break the assim tests, depending on whether Matt's tests have a fixed HISTORY rc file or rely on the HISTORY.rc file produced by ldas_setup.  We can fix this later if needed.  If so, my preference is to update the test config files (i.e., add HISTORY.rc files to the tests if they don't have it now), rather than edit the source code again. 

For now, please take a look at my changes and let me know if you see any obvious errors or issues.  Thanks!

gmao-rreichle:
> It looks right. But I will need to run the ldas_setup and compare the output. Are you going to remove 'SMAP_L4_SM_gph' for all simulation?

The idea is to have the same pre-defined collections as before, including 'SMAP_L4_SM_gph', although the latter always remains turned off when the HISTORY rc file is produced by ldas_setup from the GEOSldas_HIST.rc template.  The user can easily edit the resulting HISTORY file and turn on the 'gph' collection if desired.  

More generally, it's impossible to isolate the user completely from editing HISTORY.rc, and it makes little sense to write 'gph' when GEOSldas does assimilation and 'lnd' otherwise (which is the current behavior).   

Now that EASE tile space can be regridded to lat/lon via MAPL, we no longer need to restrict the options like we did before. Generally, I still think a simulation on the EASE tile space should write output in tile space.  But there's nothing wrong with offering the 2d lat/lon output as an option.  (Previously, uncommenting the 2d output for an EASE tile space would have resulted in garbage.)

gmao-rreichle (approved):

Issue #119, opened 2025-06-24
Is GEOSldas_LDAS.rc ignored?


weiyuan-jiang:
The default values in GEOSldas_LDAS.rc are not captured in this function [_produceExeInp](https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/dd54caaed49d86edb46183cd9845c6207fff42ac/GEOSldas_App/ldas_setup#L1636)
Is that intentional or a bug ?

@saraqzhang @gmao-rreichle

gmao-rreichle:
Not sure what exactly isn't captured.  There are no defaults for [EXP_ID, EXP_DOMAIN, ..., BCS_RESOLUTION].
Then there's the block:
https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/dd54caaed49d86edb46183cd9845c6207fff42ac/GEOSldas_App/ldas_setup#L1778
where the behavior depends on ladas_cpl.  Which case are you asking about?

weiyuan-jiang:
> > Maybe this is a bug introduced by me. The default values should be added to the output [out_dict](https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/dd54caaed49d86edb46183cd9845c6207fff42ac/GEOSldas_App/ldas_setup#L1832)
> 
> This particular line is related to GEOS_SurfaceGridComp.rc
> 
> > But somehow the lines saved here are not used. [lines](https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/dd54caaed49d86edb46183cd9845c6207fff42ac/GEOSldas_App/ldas_setup#L1790)
> 
> This particular line is related to GEOSldas_LDAS.rc
> 
> How are these two code sections connected? There may be nothing wrong here
The values in GEOSldas_LDAS.rc are not in out_dic.  So my original question still holds: Is this intentional?

gmao-rreichle:
> The values in GEOSldas_LDAS.rc are not in out_dic. So my original question still holds: Is this intentional?

Thanks, @weiyuan-jiang!  I see the issue now.  The "out_dict" argument for _ProduceExeInp() was added to create a copy of the sample exeinp file inside memory for the coupled land-atm DAS setup (i.e., avoid writing and reading back the sample exeinp file).  It is meant to capture all default values that come with the tag (including those in GEOSldas_LDAS.rc and GEOS_SurfaceGridComp.rc).  

[...]

PR #120, opened 2025-06-24
fixed default values from GEOSldas_LDAS.rc for ladas

weiyuan-jiang:


gmao-rreichle:
See https://github.com/GEOS-ESM/GEOSldas_GridComp/issues/119#issuecomment-3001816445

PR #138, opened 2025-08-11
Add reader for surface meteorological forcing from S2S-3

gmao-qliu:
We need to read S2Sv3 output for offline forecast experiments. Since S2Sv3 doesn't output 3-hourly downward shortwave and longwave radiation, we could rename the land forcing variables in post-processing. Ideally, we would extend get_GEOS() to read both S2Sv3 Forecast and Analysis outputs.

Related PRs: 
[GEOSldas PR#831](https://github.com/GEOS-ESM/GEOSldas/pull/831)

Successfully 0-diff tested for standard Intel builds after 
- ff95e4eb728bed44c14a3bba56757d1638f0ba7d (14 Aug 2025) by @gmao-rreichle 
- fe7b77109647462688b72f4db6d970dce56e3944 (20 Aug 2025) by @gmao-rreichle

### DIFF
--- a/GEOSmetforce_GridComp/LDAS_Forcing.F90
+++ b/GEOSmetforce_GridComp/LDAS_Forcing.F90
@@ -3940,9 +3940,9 @@ contains
              ! special S2S3 FCST case: must skip j==1 at S2S3 FCST initialization time because (monthly) file
              !                         exists but does not contain data for "date_time_bkwd"
              
-             write (S2S3_init_YYYYMMDD(1:4),'(i4.4)') date_time_tmp%year
-             write (S2S3_init_YYYYMMDD(5:6),'(i2.2)') date_time_tmp%month
-             write (S2S3_init_YYYYMMDD(7:8),'(i2.2)') date_time_tmp%day
+             read (S2S3_init_YYYYMMDD(1:4),'(i4.4)') date_time_tmp%year
+             read (S2S3_init_YYYYMMDD(5:6),'(i2.2)') date_time_tmp%month
+             read (S2S3_init_YYYYMMDD(7:8),'(i2.2)') date_time_tmp%day
              
              date_time_tmp%hour = 0
              date_time_tmp%min  = 0
--- a/GEOSmetforce_GridComp/LDAS_Forcing.F90
+++ b/GEOSmetforce_GridComp/LDAS_Forcing.F90
@@ -35,6 +35,7 @@ module LDAS_ForceMod
   use LDAS_DateTimeMod,                 ONLY:     &
        date_time_type,                            &
        augment_date_time,                         &
+       datetime_eq_refdatetime,                   &
        datetime_lt_refdatetime,                   &
        datetime_le_refdatetime,                   &
        is_leap_year,                              &
--- a/GEOSmetforce_GridComp/LDAS_Forcing.F90
+++ b/GEOSmetforce_GridComp/LDAS_Forcing.F90
@@ -3939,9 +3939,9 @@ contains
              ! special S2S3 FCST case: must skip j==1 at S2S3 FCST initialization time because (monthly) file
              !                         exists but does not contain data for "date_time_bkwd"
              
-             write (YYYYMMDD(1:4),'(i4.4)') date_time_tmp%year
-             write (YYYYMMDD(5:6),'(i2.2)') date_time_tmp%month
-             write (YYYYMMDD(7:8),'(i2.2)') date_time_tmp%day
+             write (S2S3_init_YYYYMMDD(1:4),'(i4.4)') date_time_tmp%year
+             write (S2S3_init_YYYYMMDD(5:6),'(i2.2)') date_time_tmp%month
+             write (S2S3_init_YYYYMMDD(7:8),'(i2.2)') date_time_tmp%day
              
              date_time_tmp%hour = 0
              date_time_tmp%min  = 0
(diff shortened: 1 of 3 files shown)


gmao-qliu:
@gmao-rreichle I updated the code. Please begin reviewing the changes. 

MET_TAG for S2Sv3  would be:
**GEOSS2S3FCST__ens?__YYYYMMDD**  for forecast forcing
and 
**GEOSS2S3AODAS**  for analysis forcing. AODAS forcing data will be stored in daily files under the usual GEOS directory structures (/diag/Y../M../)

GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -3890,8 +3969,9 @@ subroutine get_GEOS( date_time, force_dtstep, met_path, met_tag,             &
              single_time_in_file = .not. daily_met_files  ! MERRA-2 files are daily files
 
           end if
-          
-          if ( file_exists) then
+
+ 
+          if ( file_exists .and. (.not. is_S2S3_fcst) ) then      ! S2S3 FCST has monthly files
```
gmao-rreichle: @gmao-qliu :  I'm hoping this works, but I'm not 100% sure.  Could this be an issue for S2S3 AODAS forcing (at the very beginning of the AODAS product)?  Maybe `(.not. is_S2S3_fcst)` should be  `(.not. S2S3_file_specs)`??

gmao-rreichle (commented):
@gmao-qliu : I think I have something that should be close to working as desired.  Pls note the comment below.  We also still need to add documentation of the new reader in the GEOSldas repo: `doc/README.MetForcing_and_BCS.md`, see https://github.com/GEOS-ESM/GEOSldas/pull/831

PS: The revised implementation requires that the 'time_increment' (sp?) attribute is available in the nc4 file or otherwise finds its way into LDAS_GetVar().

GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -5539,6 +5635,8 @@ subroutine get_GEOS_forcing_filename(fname_full,file_exists, date_time, daily_fi
     character( 16) :: time_stamp
     character(  4) :: YYYY,  HHMM, day_dir
     character(  2) :: MM,    DD  
+    character(  8) :: S2S3_init_YYYYMMDD  ! S2S3 initial date, e.g. "20160101" 
+    character(  4) :: S2S3_ens_num        ! S2S3 ensemble member, e.g. "ens1"
```
gmao-rreichle: @gmao-qliu : For S2S3, is there a chance that there are more than 10 ensemble members with the same start date?  I forget how the "ensemble" is constructed.  I think a given month's ensemble is made up of forecasts that start on different days, with each start date having at most a few (<10) members.  If that's the case, we should be ok.  Just double-checking

gmao-qliu: The last start date of each month (i.e., approximately 1 in 6) has 10 ensemble members for longer forecasts, while the rest have 5 members. I think we should avoid this special case and stick with 5 members in all offline forecast experiments.

gmao-rreichle: For the experiments, we can avoid the 10-member case, but for the PR we need to change the "ensX" format to "ensXX".

GEOSmetforce_GridComp/LDAS_Forcing.F90
gmao-qliu: It doesn't work because when j = 2 for tmp_init, it will land in the "else" block and report an error. It should exit the loop, i.e., land in the "file_exists" block.

gmao-rreichle: @gmao-qliu : I pushed what I hope will fix this problem.  One important thing to keep in mind is that "tmp_init" refers to the initialization of the met forcing in LDAS, whereas the issue with the lack of "bkwd" forcing in S2S3 FCST data occurs only at the S2S3 FCST initialization time (S2S3_init_YYYYMMDD minus 3 hours).  The old "if" condition would have made the LDAS fail the stop/start regression test (because "tmp_init" would have triggered the "bkwd" --> "fwd" switch for every LDAS restart time (regardless of the S2S3 FCST initialization time).

GEOSmetforce_GridComp/LDAS_Forcing.F90
gmao-rreichle: @gmao-qliu:  For the case of 10 ens members on the last start date of the month, is the format of the file name substring:
1) "ens1", "ens2", ..., "ens9", "ens10"?  
2) "ens0", "ens1", ..., "ens8", "ens9"? 
3) "ens01", "ens02", ..., "ens09", "ens10"? 

I can only hope it's not case 1...

gmao-qliu: correction: the special case is 15 members not 10, and the format is case 1, i.e. "ens1", "ens2", ..."ens10", "ens11', ..

gmao-qliu:
> PS: The revised implementation requires that the 'time_increment' (sp?) attribute is available in the nc4 file or otherwise finds its way into LDAS_GetVar().

I tested this revision with "time_increment" attribute. It does work.

gmao-rreichle (approved):

PR #131, opened 2025-06-27
River routing for GEOSldas

zyj8881357:
This pull request implements the initial setup for river routing in GEOSldas, involving changes to the GEOSldas_GridComp component. It is submitted in parallel with a [similarly titled pull request in GEOSgcm_GridComp](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1023).

As an initial step, this PR makes minimal modifications:

    Updates GEOSldas_App/ldas_setup to link the river_input folder from the bcs land shared directory to the experiment's input directory.

    Adds the corresponding River_Routing_FILE entry to the LDAS.rc configuration file.

Further development and enhancements to the river routing implementation are expected in future commits.

PR #145, opened 2025-10-29
Add  river routing grid comp

weiyuan-jiang:
Add river routing module.  Replaces #131 

If river routing grid comp is added, its checkpoint should be renamed and backed up. 

Related PRs:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1143 (merged)
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1375 (merged)
https://github.com/GEOS-ESM/GEOSldas/pull/850

Testing: 
- Standard nightly 0-diff tests completed by @biljanaorescanin on 3 May 2026 (https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1375#issuecomment-4366456551)
- Two simulations (with routing) with and without land**ice** tiles were compared and found to be zero-diff for land and routing outputs by @zyj8881357 on 21 April 2026.  
- New routing functionality tested again by @zyj8881357 on 3 May 2026 (https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/145#issuecomment-4367743520).

Left for another PR: 
- enable ensemble functionality for routing (see 6ea9bbf95a12fb671548b55ca17646e7b1ede8fa)
- enable routing of runoff from landice tiles


gmao-rreichle:
@weiyuan-jiang, @zyj8881357 : When we add River Routing to GEOSlas, I would expect that we also modify the HISTORY template (GEOSldas_HIST.rc), as well as at least one config parameters that turns routing on/off (and possibly more).  Do you have a list of what is still missing from this PR?

weiyuan-jiang:
1) We need to find the permanent place for the BCS required by river-routing
2) We need this PR be approved ( https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1143  ). Maybe we should integrate make_bcs with the scripts Yujin provided to create river-routing BCs and move them to the appropriate directories
3) Change this PR accordingly

zyj8881357:
> @weiyuan-jiang, @zyj8881357 : When we add River Routing to GEOSlas, I would expect that we also modify the HISTORY template (GEOSldas_HIST.rc), as well as at least one config parameters that turns routing on/off (and possibly more). Do you have a list of what is still missing from this PR?

We already have a option called RUN_ROUTE in our LDAS.rc that turns routing on/off (1/0) which also appears in our "exe"_input_filename template. I think we can continue to use this one (that is what we did in the PR). 

If you'd like, we can add an option to turn on/off the reservoirs/lakes, but in my view, we don't need to do that because the reservoirs/lakes make the overall simulation better and we can keep less parameters in the config list.

zyj8881357:
> @weiyuan-jiang, @zyj8881357 : When we add River Routing to GEOSlas, I would expect that we also modify the HISTORY template (GEOSldas_HIST.rc), as well as at least one config parameters that turns routing on/off (and possibly more). Do you have a list of what is still missing from this PR?

In addition to what Weiyuan mentioned, we also need a place to store the raw input data for the pre-processing package that converts the raw data to the file "river_input.nc" that would be used by the make_bcs. 

But maybe this can be done later because any changes in the make_bcs and routing model may affect the pre-processing package. We can first finalize the routing_model and make_bcs, and then work on the pre-processing package.

GEOS_LdasGridComp.F90
```diff
@@ -253,6 +261,12 @@ subroutine SetServices(gc, rc)
           LANDICE(i) = MAPL_AddChild(gc, name=childname, ss=LandiceSetServices, rc=status)
           VERIFY_(status)
        endif
+
+      if (RUN_ROUTE == 1) then
+          childname='ROUTE'//trim(ensid_string)
```
gmao-rreichle, line 270: @weiyuan-jiang, @zyj8881357 : This line suggests that the routing model could be set up with LDAS in ensemble mode.  The same seems to be true for the landice model (see preceding block).  However, [the ensemble GridComp](https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/develop/GEOSens_GridComp/GEOS_EnsGridComp.F90) appears to be exclusively in the (non-glaciated) **land** tile space.  Since the land, landice, and routing GridComps each work in different tile spaces, do we need to overhaul the ensemble GridComp to make the ensemble functionality work for routing and/or landice?  I wonder how long it might take to do that.  In the meantime, we could perhaps add a "stop" if `NUM_ENSEMBLE>1 .and. (with_landice .or. RUN_ROUTE>0)`; that is, turn off the ensemble functionality unless the simulation includes only land tiles.  Please advise, thanks

zyj8881357: "stop if NUM_ENSEMBLE>1 .and. (with_landice .or. RUN_ROUTE>0)" has been added.

weiyuan-jiang:
This line will be changed
https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/211cb1de686c878b85460fa8e232080fa67274e4/GEOSldas_App/ldas.py#L1007

gmao-rreichle:
@zyj8881357, @weiyuan-jiang : The standard routing configuration in GEOSldas uses a 1-hour time step for routing (`RRM_DT: 3600` in GEOS_SurfaceGridComp.rc) and a heartbeat of 7.5 minutes.  Do we have the following "assert" statements somewhere in the routing GC or in LDAS? 
- assert that RRM_DT is a multiple of HEARTBEAT
- assert that segment length >= RRM_DT  (a simulation with routing must complete at least one RRM_DT time step)

There may be other requirements that need to be verified.

weiyuan-jiang:
> This line will be changed
> 
> https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/211cb1de686c878b85460fa8e232080fa67274e4/GEOSldas_App/ldas.py#L1007

@zyj8881357 Is the line ready for change?

zyj8881357:
> > This line will be changed
> > https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/211cb1de686c878b85460fa8e232080fa67274e4/GEOSldas_App/ldas.py#L1007
> 
> @zyj8881357 Is the line ready for change?

@weiyuan-jiang Yes. The path is /discover/nobackup/projects/gmao/bcs_shared/restarts/surface/route/v202601/ . Could you please help me to update it in the code? I am not familiar with the string variable for dir "/discover/nobackup/projects/gmao/bcs_shared/" in the script. Thank you!

weiyuan-jiang:
> > > This line will be changed
> > > https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/211cb1de686c878b85460fa8e232080fa67274e4/GEOSldas_App/ldas.py#L1007
> > 
> > 
> > @zyj8881357 Is the line ready for change?
> 
> @weiyuan-jiang Yes. The path is /discover/nobackup/projects/gmao/bcs_shared/restarts/surface/route/v202601/ . Could you please help me to update it in the code? I am not familiar with the string variable for dir "/discover/nobackup/projects/gmao/bcs_shared/" in the script. Thank you!

I realized that we don't support RESTART=0 any more. I removed it and gave users a hint

GEOS_LdasGridComp.F90
gmao-rreichle: @zyj8881357, @weiyuan-jiang : Do we now know whether routing is supported in ensemble mode (NUM_ENSEMBLE>1)?

GEOSldas_App/ldas.py
```diff
@@ -32,10 +32,11 @@ def __init__(self, cmdLineArgs):
         """
         """
         # These keywords are excluded from LDAS.rc (i.e., only needed in pre- or post-processing)
+        # or their values are re-assigned in LDAS.rc 
         self.NoneLDASrcKeys=['EXP_ID', 'EXP_DOMAIN',
             'BEG_DATE', 'END_DATE','RESTART','RESTART_PATH',
             'RESTART_DOMAIN','RESTART_ID','BCS_PATH','TILING_FILE','GRN_FILE','LAI_FILE','LNFM_FILE','NIRDF_FILE',
-            'VISDF_FILE','CATCH_DEF_FILE','NDVI_FILE',
+            'VISDF_FILE','CATCH_DEF_FILE','NDVI_FILE', 'TILINGNC4_FILE','TILE2PFAF_FILE',
```
gmao-rreichle: @weiyuan-jiang : I can't quite figure out if we still need `'TILINGNC4_FILE','TILE2PFAF_FILE'` here.  We certainly don't have a separate "tile2pfaf.nc" file anymore.  Is the config variable `'TILE2PFAF_FILE'` now used to point to the "Pfafstetter" tile file?  And do we need  `'TILINGNC4_FILE'`?
My sense is that we don't want these config variables in the exeinp file because we should be able to derive them from the bcs info that is already provided in the exeinp file. 
I understand that the two config variables need to be provided to [GEOS_RouteGridComp (Lines 619-621)](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/200a949903aff011170f8875b54311ea184c2a88/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSroute_GridComp/GEOS_RouteGridComp.F90#L619-L621):

```
        call MAPL_GetResource (MAPL, tile_pfaf_file, label = 'TILE2PFAF_FILE:',  default = '../input/route_tile.nc4', RC=STATUS ) 
        call MAPL_GetResource (MAPL, tile_file,      label = 'TILINGNC4_FILE:',     default = '../input/tile.nc4',       RC=STATUS )          
```

But the comments above describing `self.NoneLDASrcKeys` suggest that the config variables in this list are only needed in pre- or post-processing. 

So how do we communicate these config variables to GEOS_RouteGridComp?  And what (if anything) needs to be done in ldas_setup to make this work?  One thing that may be still be needed is the "translation" of the old "SMAP-EASEvx-Mxx" grid name to "EASEvx_Mxx" (see Line ~360 below)

gmao-rreichle (commented):
@weiyuan-jiang @zyj8881357 , see comments below

GEOSldas_App/ldas.py
```diff
@@ -1243,18 +1334,12 @@ def createRCFiles(self):
                 for key,val in zip(rstkey,rstval) :
                     keyn = key+ '_INTERNAL_RESTART_FILE'
                     valn = '../input/restart/'+val+tmpl_+'_internal_rst'
-                    ldasrcInp[keyn]= valn
+                    ldasrcInp[keyn] = valn
+                    if 'VEGDYN' not in key: # not checkpoi for vegdyn
+                       keyn = key + '_INTERNAL_CHECKPOINT_FILE'
+                       valn = val + tmpl_ + '_internal_checkpoint'
```
gmao-rreichle, line 1331: @weiyuan-jiang : These three lines replace the following:

```
                if self.with_land :
                   keyn = catch_ + '_INTERNAL_CHECKPOINT_FILE'
                   valn = self.catch+tmpl_+'_internal_checkpoint'
                       ldasrcInp[keyn] = valn

                if self.with_landice :
                   keyn = 'LANDICE_INTERNAL_CHECKPOINT_FILE'
                   valn = 'landice'+tmpl_+'_internal_checkpoint'
                   ldasrcInp[keyn]= valn

```

Is this change intentional, or maybe the result of a merge error?  It's not obvious to me that the "if" conditions in the old code are no longer needed.   It may all be ok, but I thought I should flag it for review.  Thanks

GEOS_LdasGridComp.F90
weiyuan-jiang: I think this is a general approach without adding too many "if else". It doesn't hint it supports ensemble now. It may support in the future

GEOSldas_App/ldas.py
weiyuan-jiang: it is intentional. I actually should simplify the code further

GEOSldas_App/ldas.py
weiyuan-jiang: TILE2PFAF_FILE is the "Pfafstetter" tile file for EASE grid. I think we don't need this "TILINGNC4_FILE". It is the "TILE_FILE"

gmao-rreichle: Ok, thanks.  I'll try to think of a clearer name, although not sure that's possible.  If nothing comes to mind, I'll add some comments for clarification.
Re. TILINGNC4_FILE:   The routing calcs in the GCM GC may be using the nc4 tile file, whereas the TILE_FILE is probably the ASCII version.  The only difference in the file name should be the extension (".til" vs. ".nc4").  There shouldn't be a need for two distinct config variables, but we do need to figure out how to handle the file name extension

zyj8881357: > Ok, thanks. I'll try to think of a clearer name, although not sure that's possible. If nothing comes to mind, I'll add some comments for clarification. Re. TILINGNC4_FILE: The routing calcs in the GCM GC may be using the nc4 tile file, whereas the TILE_FILE is probably the ASCII version. The only difference in the file name should be the extension (".til" vs. ".nc4"). There shouldn't be a need for two distinct config variables, but we do need to figure out how to handle the file name extension

Yes, the routing model reads the nc4 version of tile file while the default version is still .til. So that is why I set the TILINGNC4_FILE here.

weiyuan-jiang: We don't need this extra name. We can test if this file a ASCII or nc4 , then read it with different method.

zyj8881357:
> * Compare simulation (with routing) that has only land tiles with another simulation (with routing) that also has landice tiles (in addition to land tiles).  The land and routing outputs for the two simulations should be zero-diff.

I have done this test. The runs with and without landice tiles are zero-diff for land and route.

GEOS_LdasGridComp.F90
gmao-rreichle: My commit 6ea9bbf95a12fb671548b55ca17646e7b1ede8fa undoes @weiyuan-jiang's earlier commit 45304d2cc5b50a60fc2e139126670ac71835468c, which was an attempt to provide ensemble functionality but which did not work (because the routing and land tile spaces are different).  
Note that in GEOS_LdasGridComp.F90 we already have a protection against running in ensemble mode when routing is turned on.

gmao-rreichle:
@zyj8881357  @biljanaorescanin
I think the current routing PRs are now ready for final testing:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1375
https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/145   (this PR)
https://github.com/GEOS-ESM/GEOSldas/pull/850

@biljanaorescanin : When you get a chance, please run the usual 0-diff tests for LDAS and the GCM

@zyj8881357 : Please do one more check of a routing simulation with the latest GEOSgcm_GridComp and GEOSldas_GridComp branches. 

cc: @weiyuan-jiang

zyj8881357:
> @zyj8881357 @biljanaorescanin I think the current routing PRs are now ready for final testing: [GEOS-ESM/GEOSgcm_GridComp#1375](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1375) #145 (this PR) [GEOS-ESM/GEOSldas#850](https://github.com/GEOS-ESM/GEOSldas/pull/850)
> 
> @biljanaorescanin : When you get a chance, please run the usual 0-diff tests for LDAS and the GCM
> 
> @zyj8881357 : Please do one more check of a routing simulation with the latest GEOSgcm_GridComp and GEOSldas_GridComp branches.
> 
> cc: @weiyuan-jiang

@gmao-rreichle I have finished both the zero-diff and scientific test with the latest GEOSgcm_GridComp and GEOSldas_GridComp branches. They work well and can produce reasonable discharge output against with the observations.

gmao-rreichle (approved):

PR #146, opened 2025-11-07
enable remapping of landice restarts (from CF to EASE) via ldas_setup

biljanaorescanin:
@weiyuan-jiang created a fix so we can remap landice restart files from `CF` to `EASE` tile space via `ldas_setup`. 

More details: 
Currently, this is the only way to obtain a starting point for `Land-Ice` `EASE` grid simulations with `GEOSldas`. It may actually be a better approach, since it allows the use of any `CF` restart file.
The `RESTART:M` option is not working because the MERRA2 restart files have been moved from the archive. This issue should be resolved eventually, with only days `14` and `28` available as options. Until then, this method is required to initialize EASE grid Land-Ice runs.

Successfully 0-diff tested by @gmao-rreichle on 25 Nov 2025


biljanaorescanin:
For testing 1day global case run was done from C90 to EASE M36 using (`model_tests_with_landice/HISTORY_1d.rc`).

biljanaorescanin:
`Let me sum current issue we have here for records: `

It looks like land-ice group plan is to run land-ice only experiments (tile type 20). And then use that restart to remap and start new exp on diff grid? 

`In that configuration we cannot use ldas_setup with RESTART:2 to remap, because:`

ldas_setup expects both tile types 100 (land) and 20 (land-ice) to be present in the input restarts in order to create a new restart on a different grid.

If I follow Devon’s example and try to run ldas_setup with just a type-20 restart on the M09 EASE grid, preprocessing completes (catparam/domain are created), but no mapping file or restart is created, and ldas_setup eventually crashes with a TypeError: expected str, bytes or os.PathLike object, not NoneType when it tries to find the input restart path (self.in_rstfile is None because the restart wasn’t created).

`So the practical implication is:`

To remap from C1440 to something like M09 using ldas_setup as currently implemented, we need a run that includes both tile types 100 and 20, so that both land and land-ice restarts exist at the end of the run.

After ldas_setup is used with those combined restarts, you can still run only land-ice going forward if you want the dual tile requirement is just for the restart/remap step.

**Alternatively, ldas_setup itself would need to be modified to support starting from land-ice only restarts.**

`Bottom line: with the current code, the existing land-ice only experiments (type 20 only) cannot be used as RESTART:2 sources for remapping; we need at least one spinup with both tile types or a change to ldas_setup.`

@weiyuan-jiang this is just a summary of what we found on from Teams chat. Where starting point was Devon's run.

weiyuan-jiang:
@lcandre2 , Now this branch should support setup with only landice

gmao-rreichle (approved):

PR #147, opened 2025-11-13
Add start/stop model regression testing package

biljanaorescanin:
Add regression test framework for model runs

This commit introduces a new regression testing framework under
`GEOSldas_App/util/postproc/regression` .

It includes:
- `README.md` with detailed instructions for running regression tests
- `start_stop_model.sh` script to manage model execution for tests
- `HISTORY_1d.rc` and `HISTORY_2d.rc` templates to ensure consistent
  output (for both CS and EASE grids)
- Updates to `ldas_setup` to enable regression test configuration

Notes:
1. At this stage, the framework supports **model-only** regression runs
   (not assimilation).
2. Depending on tile types, the model may run land-only (lfs, lnd) or land+landice (lfs, lnd, glc)  collections. The regression now detects this automatically.
4. This work addresses one of the items discussed in:
   https://github.com/GEOS-ESM/GEOSldas/issues/250
5. The design follows the approach used in GEOS GCM regression testing.
   Since GEOSldas does not support 2D layout tests, we use a 1D-invariant
   approach analogous to:
   https://github.com/GEOS-ESM/GEOSgcm_App/wiki/Regress-Script-Flow-Chart
6. This implementation is a proposed starting point. The approach can
   be modified or the PR closed if it is not aligned with how we wish to
   support regression testing in GEOSldas.

### DIFF
--- a/GEOSldas_App/ldas_setup
+++ b/GEOSldas_App/ldas_setup
@@ -3,8 +3,10 @@
 import sys
 import argparse
 import resource
+import os, shutil
 from setup_utils import *
 from ldas        import *
+from pathlib     import Path
 
 def parseCmdLine():
     """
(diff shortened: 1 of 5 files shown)


biljanaorescanin:
`Testing summary: `
Trivially zero diff for GEOSldas nightly regression tests. 

`Regress runs tested: `
`CF90 tile type 100 20 ` (collections lnd, lfs and glc)
`CF90 tile type 100` (collections lnd and lfs )
`EASE M36 tile type 100 20`
`EASE M36 tile type 100`

biljanaorescanin:
After discussion, we’ve decided to pause this effort due to the expected maintenance overhead. For now, regression testing will remain non-automated, but we can revisit this approach in the future if needed.

PR #148, opened 2025-11-14
fix bug in ASCAT_EUMET soil moisture obs reader

gmao-rreichle:
A bug was discovered in the ASCAT_EUMET soil moisture obs reader. 
In a nutshell, after reading the obs from the files that overlap with the assimilation window, the obs were not properly screened for falling into the assimilation window.  This resulted in ~40% of the observations being assimilation twice.

The fix is trivially 0-diff only in the sense that none of the standard tests engage ASCAT soil moisture assimilation. 

For simulations with ASCAT soil moisture assimilation, the fix is naturally **not** 0-diff.

Successfully 0-diff tested by @gmao-rreichle, 14 Nov 2025

@amfox37 @gmao-qliu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed bug in ASCAT EUMET soil moisture obs reader.
 - Fixed Restart=1 when the domain is not global.
 
 ### Removed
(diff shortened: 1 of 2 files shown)


gmao-rreichle:
@amfox37 : When you get a chance, please verify that the fix is correct and let me know.  Thanks!

amfox37 (commented):
Yes, I believe this is all that needs changing. We're conducting quite extensive testing of its impact, of course. 
I also checked similar logic elsewhere in the ASCAT reader and for other obs readers - SMAP, SMOS, CYGNSS - and didn't see any issues there.

biljanaorescanin (approved):

PR #160, opened 2026-02-11
revised RESTART options;  fix string matching for EASE tile file 

weiyuan-jiang:
Revised and cleaned up GEOSldas RESTART options:
- Clarified scope and constraints of RESTART=1 and RESTART=2.  
- For RESTART=1, the tile file and bcs files are now pulled from the GEOSldas experiment in RESTART_PATH, with a basic check for bcs consistency between the restart experiment and the new experiment. 
- Introduced RESTART=3 for restarting from restarts produced by the GCM.
- Cleaned up stray code from RESTART=G, which had been removed some time ago. 

Fixed string matching for EASE tile file to accommodate "EASE*-Pfafstetter" tile file that is added into bcs directory for the benefit of runoff routing (https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1143)

Testing:
- Successfully zero-diff tested by @gmao-rreichle on 18 March 2026;  zero-diff result required replacing inconsistent bcs files (greenness, NIRDF, VISDF) in the RESTART_PATH in Matt's test setup.  These inconsistent bcs files were not used in the old version of the code but are used after the changes implemented in the PR. 
- RESTART='2' and RESTART='3' are not used in nightly zero-diff tests and were tested separately by @biljanaorescanin on 19 March 2026 https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/160#issuecomment-4091819948


gmao-rreichle:
@weiyuan-jiang : Thanks for adding this bug fix.  One question: Do we also need the same protection here (for 'RESTART'==2)?
https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/68d08a1e10bb22782137fa0d852b1698e2a31520/GEOSldas_App/ldas.py#L376

It's a bit unclear to me why we handle the TILING_FILE for 'RESTART'==1 in a different place in the code from where we handle the TILING_FILE for 'RESTART'==2.   In both cases, we have restart files (and thus a tiling file) from an existing GEOSldas experiment.   Shouldn't the logic for obtaining the TILING_FILE (from the ./rc_out directory) be the same (or nearly the same) in both cases?   (For 'RESTART'==2, I think we perface the tiling file name with 'MAPL_', but that could be done with an if-condition that applies to just the file name within the larger logic of assembling the path and deciding if we use the txt or nc4 file.   

Also, I wonder if the code involved in the bug fix here also is related to the bug that Devon reported.  It may make sense to fix both bugs in this PR (if Devon's problem is in fact a bug in GEOSldas setup).

GEOSldas_App/ldas.py
```diff
@@ -296,6 +296,8 @@ def __init__(self, cmdLineArgs):
            if 'MAPL_' in os.path.basename(f):
               txt_tile = [f]
               break
+        nc4_tmp  = glob.glob(inpgeom_ + '*.nc4' + domain_)
+        nc4_tile = [ item_ for item_ in nc4_tmp if 'tile2pfaf' not in item_ ]
         nc4_tile = glob.glob(inpgeom_ + '*.nc4' + domain_)
```
gmao-qliu, line 325: need to remove line 301

weiyuan-jiang: Thanks. Removed

weiyuan-jiang:
> @weiyuan-jiang : Thanks for adding this bug fix. One question: Do we also need the same protection here (for 'RESTART'==2)?
> 
> https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/68d08a1e10bb22782137fa0d852b1698e2a31520/GEOSldas_App/ldas.py#L376
> 
> It's a bit unclear to me why we handle the TILING_FILE for 'RESTART'==1 in a different place in the code from where we handle the TILING_FILE for 'RESTART'==2. In both cases, we have restart files (and thus a tiling file) from an existing GEOSldas experiment. Shouldn't the logic for obtaining the TILING_FILE (from the ./rc_out directory) be the same (or nearly the same) in both cases? (For 'RESTART'==2, I think we perface the tiling file name with 'MAPL_', but that could be done with an if-condition that applies to just the file name within the larger logic of assembling the path and deciding if we use the txt or nc4 file.
> 
> Also, I wonder if the code involved in the bug fix here also is related to the bug that Devon reported. It may make sense to fix both bugs in this PR (if Devon's problem is in fact a bug in GEOSldas setup).

This is not a bug for Devon because she did not setup run on EASE grid. I will make changes on the route PR when we need the tile2pfaf file in the rc_out directory. But I agree it is quite messy here. I will do some cleanup in the other PR

gmao-rreichle:
> This is not a bug for Devon because she did not setup run on EASE grid. I will make changes on the route PR when we need the tile2pfaf file in the rc_out directory. But I agree it is quite messy here. I will do some cleanup in the other PR

My comment wasn't meant to be specific to the EASE grid.   I think the code in question executes independent of the grid/tile space.  Is there a reason for the separation of what seems to be (nearly) the same action into a block for RESTART==1 and another block elsewhere  for RESTART==2?
  
Re. Devon, I was just wondering if the TILING file is processed correctly.   The bug fix here touches the lines that are dealing with the nc4 tiling file, and we suspected that Devon's problem may have to do with a mismatched TILING file.

In any case, the bug fix that the PR intends to fix is for EASE and touches LDAS setup.  My suggestion was to add the bug fix for Devon (if there is indeed a bug) into this PR -- assuming the fix also touches LDAS setup.  (We can rename the PR to something like "LDAS setup bug fixes...")

weiyuan-jiang:
In the first Restart == 1, it actually tries to figure out where are the BC files for the outputs. It there is no 'domain', all BC files are from BC.  In the second Restart == 2, it gets input tile file for remapping later on

GEOSldas_App/ldas.py
```diff
@@ -296,7 +312,8 @@ def __init__(self, cmdLineArgs):
            if 'MAPL_' in os.path.basename(f):
               txt_tile = [f]
               break
-        nc4_tile = glob.glob(inpgeom_ + '*.nc4' + domain_)
+        nc4_tmp  = glob.glob(inpgeom_ + '*.nc4' + domain_)
+        nc4_tile = [ item_ for item_ in nc4_tmp if 'tile2pfaf' not in item_ ]
```
gmao-rreichle: @weiyuan-jiang : If we add an nc4 version of the cube-sphere-like tile file (instead of Yujin's "tile2pfaf" file, this bug fix no longer works.  Probably the string 'tile2pfaf' would need to be replaced with 'Pfafstetter'.

GEOSldas_App/ldas.py
```diff
@@ -279,15 +274,36 @@ def __init__(self, cmdLineArgs):
...
                     self.ExeInputs['RESTART_DOMAIN'], 'rc_out/'])
+        if RESTART_str == '1' :
+           inpdir_  = inp_
+           inpgeom_ = inp_
+           BCS_txt  = glob.glob(inp_ + 'BCS_info.txt')
+           if len(BCS_txt)== 0:
+              print("Warning: BCS_info.txt not found for restart experiment. User is responsible for ensuring consistency of restart and experiment BCS.")
+           elif len(BCS_txt)== 1:
+              BCS_tmp = parseInputFile(BCS_txt[0])
+              assert self.ExeInputs['BCS_PATH']       == BCS_tmp['BCS_PATH'],       "BCS_PATH does not match path from restart dir ("             + BCS_tmp['BCS_PATH']       + ")"
```
gmao-rreichle: @weiyuan-jiang : It's a bit more involved, but can we distinguish the VERSION from the path to the bcs?  For offline simulations, the trailing dir name in the path is the bcs VERSION.  It would be better to limit the check to just VERSION. This avoids triggering assert.false if the difference is only in the path (e.g., /discover/nobackup/projects vs /gpfsm/dnb05/projects.

gmao-rreichle (commented):
@weiyuan-jiang : I added a couple of commits and the following comments.  Please take a look and let me know if you see anything wrong or have any questions, thanks

GEOSldas_App/ldas.py
```diff
@@ -346,7 +376,8 @@ def __init__(self, cmdLineArgs):
 
         for key,val in _domain_dic.items() :
             if key in self.ExeInputs :
-                _domain_dic[key]= self.ExeInputs[key]
+               _domain_dic[key]= self.ExeInputs[key]
+               assert RESTART_str != 2, "Please comment out " + key + " event you set it global"
```
gmao-rreichle: @weiyuan-jiang : This new line may need work:
- only RESTART=1 allows zooming in, so a check for the domain to be global should apply to RESTART=2, 3, and M, not just RESTART=2
- It's possible that a user has _domain_dic[key] in the exe input file with values that specify a global domain.  In this case, we should not exit. If I understand the logic correctly, however, the "assert" here would exit simply based on the presence of the _domain_dic[key] strings in the exe input file, regardless of the values.

There is already a check for the domain being global for RESTART=2 earlier (Line ~306):

```
 elif RESTART_str == '2':
           txt_tile = glob.glob(inp_ + '*.domain')
           assert len(txt_tile) == 0, "For RESTART=2, must restart from and run on global domain"
```

I think we can drop the "assert" here.  Not sure if we could find a better way to check for the domain being global for RESTART=2, 3, and M

GEOSldas_App/ldas.py
```diff
@@ -279,15 +275,45 @@ def __init__(self, cmdLineArgs):
...
+           assert len(txt_tile) == 0, "For RESTART=2, must restart from and run on global domain"
+           in_tilefiles_ = glob.glob(inp_+'MAPL_*.til')
+           if len(in_tilefiles_) == 0 :
+              nc4_tmp = glob.glob(inp_+'/*.nc4')
+           if len(in_tilefiles_) == 0 :
+              in_tilefiles_ = glob.glob(inp_+'/*.til')
+           # for EASE tile space, pick "standard" tile file (one tile per grid cell) and not "-Pfafstetter" tile file
+           if 'EASEv' in in_tilefiles_[0]:
+              in_tilefiles_ = [ item_ for item_ in in_tilefiles_ if '-Pfafstetter' not in item_ ]
+           self.in_tilefile =os.path.realpath(in_tilefiles_[0])
```
gmao-rreichle, line 315: @weiyuan-jiang : I changed "in_tilesfile_" here to "in_tilefiles_" because it looked very much like a typo.  I did not try to run before or after my change.  Please double check that all is as it should be.  Thanks

gmao-rreichle (commented):
@weiyuan-jiang : See comments below.  After we address them, the PR would be ready for testing by @biljanaorescanin

GEOSldas_App/ldas.py
weiyuan-jiang: Let's drop the assert here. It is users' responsibility to check the domains.

GEOSldas_App/ldas.py
weiyuan-jiang: Thanks. Indeed it is typo and a potential bug

biljanaorescanin:
`Testing summary:`
1. Nightly tests pass. 
2. Option RESTART2 works. 
3. Option RESTART3 works 
4. Option M doesn't work because M2 is not on archive it changed path.

gmao-rreichle (approved):

PR #163, opened 2026-03-12
Add optional ObsFcstAna NetCDF4 output with structured species metadata

amfox37:
This PR adds end-to-end optional NetCDF4 output for ObsFcstAna (BIN / NC4 / BOTH) to address the limitations of the binary files. 

The NC4 output includes per-observation fields as-per the binary files, adds run/time metadata, and introduces a structured n_species metadata table (species_id, species_assim, species_scale, species_errstd, species_varname, species_units, species_fcstvarname, species_fcstunits, species_descr).  It also provides global attributes "species_####_descr" for quick inspection.

The PR includes the following changes in the ensupd namelist file:
- out_ObsFcstAna is changed from "logical" to "integer".
- obs_param%fcstvarname and obs_param%fcstunits are added to provide units for the Obs, Fcst, and Ana data in the ObsFcstAna file.  These new fields should not be filled by the user.  Instead, they are populated by subroutine read_ens_upd_inputs().

Testing summary: 
- Successfully 0-diff tested by @gmao-rreichle on 1 May 2026.  Interface change in nml file required test runs to be constructed manually from official test run config.  Official test config updated after successful 0-diff test on 1 May 2026.
- New functionality tested by @amfox37 (~29 Apr 2026)


GEOSldas_App/lenkf_j_template.py
```diff
@@ -481,6 +481,16 @@
...
 
+   set ObsFcses = `ls *.ldas_ObsFcstAna.*.nc4`
+   foreach obsfcs ( $ObsFcses )
+      set ThisTime = `echo $obsfcs | rev | cut -d'.' -f2 | rev`
+      set TY = `echo $ThisTime | cut -c1-4`
+      set TM = `echo $ThisTime | cut -c5-6`
+      set THISDIR = $EXPDIR/output/$EXPDOMAIN/ana/ens_avg/Y${{TY}}/M${{TM}}/
+      if (! -e $THISDIR            ) mkdir -p $THISDIR
+      /bin/mv $obsfcs ${{THISDIR}}$obsfcs
+   end
```
gmao-rreichle: This block appears to be a carbon-copy of the preceding block, with the only difference being the file name extension in the file name template used to list (ls) the files.  
I'm not sure if we really want to support output of both "bin" and "nc4" ObsFcstAna files.  But if we do, I think simply removing the file name extension from the file name template in the "set ObsFcses = " in the previous block would do the trick. Then all "ldas_ObsFcstAna" files would be moved together, regardless of the file format, using just one block of code.

GEOSldas_App/LDASsa_DEFAULT_inputs_ensupd.nml
```diff
@@ -44,6 +44,8 @@ update_type = 0
 
 out_obslog      = .true.
 out_ObsFcstAna  = .false.
+! out_ObsFcstAna_format: 'BIN' (legacy unformatted), 'NC4', or 'BOTH'
+out_ObsFcstAna_format = 'BIN'
```
gmao-rreichle: I'm wondering if there's a simpler way to configure the ObsFcstAna output with the new format option. 
If this was an entirely new set of output files, I would recommend using an integer config parameter, essentially what is now "out_ObsFcstAna_mode" in the guts of the code, with one tweak: A value of 0 would indicate that ObsFcstAna files are **not** written.  This new integer variable would replace the current logical "out_ObsFcstAna", and it would make the processing of the config much simpler. 

The problem with the above suggestion is that the ensupd nml files would not be backward compatible.  Although backward compatibility would only work if the new variable "out_ObsFcstAna_format" is optional.  Not sure if that's the case now.  

Having said that, I also wonder if we want to support the option to output "BOTH" formats.  It's certainly useful for debugging this PR, but beyond that, I don't think we can afford to write files in both formats.  

My sense is that we might need a clean break here and change the existing config variable to "integer" as follows:
out_ObsFcstAna = 0     ! no output
out_ObsFcstAna = 1      ! bin
out_ObsFcstAna = 2     ! nc4

This needs a bit more thought and deliberation.

GEOSlandpert_GridComp/random_fields.F90
```diff
@@ -397,14 +395,13 @@ subroutine rfg2d_fft(this, rseed, rfield, rfield2, lx, ly, dx, dy)
...
     real :: N_xy_fft_real
-    integer :: n1, n2
 #ifdef MKL_AVAILABLE
     integer :: status
     complex, allocatable :: z_inout(:)
     complex, pointer :: tmp_field(:,:)
     complex, pointer :: tmp_field_dim1(:,:)
     complex, pointer :: tmp_field_dim2(:,:)
-    integer :: npes, rank, ldim1, ldim2, ierror
+    integer :: n1, n2, npes, rank, ldim1, ldim2, ierror
```
gmao-rreichle: I think something went wrong when I pressed the "update current branch" button on the PR landing page.  It looks like #160 was not merged correctly.  I'm writing this comment to remind myself that this needs revisiting.

GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -340,6 +349,23 @@ subroutine read_ens_upd_inputs(               &
...
+       out_ObsFcstAna_mode = OBSFCSTANA_FMT_NC4
+
+    case ('BOTH')
+       out_ObsFcstAna_mode = OBSFCSTANA_FMT_BOTH
+
+    case default
+       err_msg = 'unknown value of "out_ObsFcstAna_format": "' // &
+            trim(out_ObsFcstAna_format) // '"'
+       call ldas_abort(LDAS_GENERIC_ERROR, Iam, err_msg)
+    end select
```
gmao-rreichle: This block seems to match what's done in Lines ~1705-1719 of clsm_ensupd_enkf_update.F90, suggesting some redundancy.  If we end up changing the config variable, these blocks probably become obsolete.  This comment serves as a reminder to double-check and edit if needed

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+       end do
+
+       call nc4_check( nf90_put_var(ncid, species_id_varid,       obs_param(1:N_obs_param)%species) )
+       call nc4_check( nf90_put_var(ncid, species_assim_varid,    species_assim_int) )
+       call nc4_check( nf90_put_var(ncid, species_scale_varid,    species_scale_int) )
+       call nc4_check( nf90_put_var(ncid, species_getinnov_varid, species_getinnov_int) )
+       call nc4_check( nf90_put_var(ncid, species_errstd_varid,   species_errstd_r4) )
+       call nc4_check( nf90_put_var(ncid, species_varname_varid,  species_varname_c40) )
+       call nc4_check( nf90_put_var(ncid, species_units_varid,    species_units_c40) )
+       call nc4_check( nf90_put_var(ncid, species_descr_varid,    species_descr_c40) )
```
gmao-rreichle: These variables seemed mangled when I read them from the sample nc4 files in Matlab.  It's probably just a transpose error in the 2d character array.   I wonder if the 2d char array is the best option for writing these variables.  Not sure if we can add the "species_" (or "obsparam_") info as Attributes, and if that would make things any simpler.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+    character(len=*), parameter :: Iam = 'write_ObsFcstAna_nc4'
+    character(len=400)          :: err_msg
+
+    call nc4_check( nf90_create(trim(fname), nf90_clobber + NF90_NETCDF4, ncid) )
+
+    call nc4_check( nf90_def_dim(ncid, 'nobs', NF90_UNLIMITED, nobs_dimid) )
+    call nc4_check( nf90_def_dim(ncid, 'n_species', N_obs_param, nspecies_dimid) )
+    call nc4_check( nf90_def_dim(ncid, 'species_strlen', 40, species_strlen_dimid) )
+
+    call nc4_check( nf90_def_var(ncid, 'assim',   NF90_INT,   [nobs_dimid], assim_varid) )
```
gmao-rreichle: Do we need to convert "assim" from logical to integer when it's written into the nc4 file?  Not sure if nc4 allows logicals.  It's probably fine to convert, but just flagging it for eventual input by the SI team.

gmao-rreichle: Making a note that I still need to review the variable names in the nc4 file.  Here, maybe "assim_flag" would be better than just "assim".  The field names in the obs structure of the source code were not meant for general data product users.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+       deallocate(species_scale_int)
+       deallocate(species_getinnov_int)
+       deallocate(species_errstd_r4)
+       deallocate(species_varname_c40)
+       deallocate(species_units_c40)
+       deallocate(species_descr_c40)
+
+    end if
+
+    call nc4_check( nf90_close(ncid) )
```
gmao-rreichle, line 2039: We're probably not compressing the nc4 files, but it may be useful to do that either when they're written or in (optional) postprocessing.  Having said that, we should check with the SI team if the nc4 stuff we are doing here may be better done through MAPL.  I don't have a preference.   I just don't know what's best.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'update_type', update_type) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'dtstep_assim', dtstep_assim) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'grid_name', trim(grid_name)) )
+
+    do i=1,N_obs_param
+       write(attr_name, '(A,I4.4,A)') 'species_', obs_param(i)%species, '_descr'
+       call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, trim(attr_name), trim(obs_param(i)%descr)) )
+    end do
+
+    call nc4_check( nf90_put_att(ncid, assim_varid, 'long_name', 'assimilation flag') )
```
gmao-rreichle: This comment serves as a reminder for me to review the LONG_NAME attributes.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+    call nc4_check( nf90_def_var(ncid, 'anavar',  NF90_FLOAT, [nobs_dimid], anavar_varid) )
+
+    call nc4_check( nf90_def_var(ncid, 'species_id',       NF90_INT,   [nspecies_dimid], species_id_varid) )
+    call nc4_check( nf90_def_var(ncid, 'species_assim',    NF90_INT,   [nspecies_dimid], species_assim_varid) )
+    call nc4_check( nf90_def_var(ncid, 'species_scale',    NF90_INT,   [nspecies_dimid], species_scale_varid) )
+    call nc4_check( nf90_def_var(ncid, 'species_getinnov', NF90_INT,   [nspecies_dimid], species_getinnov_varid) )
+    call nc4_check( nf90_def_var(ncid, 'species_errstd',   NF90_FLOAT, [nspecies_dimid], species_errstd_varid) )
+    call nc4_check( nf90_def_var(ncid, 'species_varname',  NF90_CHAR,  [species_strlen_dimid, nspecies_dimid], species_varname_varid) )
+    call nc4_check( nf90_def_var(ncid, 'species_units',    NF90_CHAR,  [species_strlen_dimid, nspecies_dimid], species_units_varid) )
+    call nc4_check( nf90_def_var(ncid, 'species_descr',    NF90_CHAR,  [species_strlen_dimid, nspecies_dimid], species_descr_varid) )
```
gmao-rreichle: Maybe we should change the "species_" prefix for the variables from the "obsparam" structure to something that includes "obs" and/or "param".  The term "species" by itself may be confusing.  Maybe "obsspecies_", or even just "obsparam_".  The key is to express that this is about parameters for obs.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'schema_version', 'ObsFcstAna_nc4_v1') )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'N_obsf',  N_obsf) )
```
gmao-rreichle: Not sure we want the trailing "f" here, which is going to be confusing to the user.  Also, isn't this the same as the dimension "nobs" ?   If yes, we probably don't need the Attribute, or at least we need to make the names consistent

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'N_obs_param', N_obs_param) )
```
gmao-rreichle: Make the name "N_obs_param" consistent with whatever we use as prefix for the variables from the "obsparam" structure.  If, say, we choose the prefix "obsspecies_", then we probably want to call this "N_obsspecies".

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'year',    date_time%year) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'month',   date_time%month) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'day',     date_time%day) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'hour',    date_time%hour) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'minute',  date_time%min) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'second',  date_time%sec) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'dofyr',   date_time%dofyr) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'pentad',  date_time%pentad) )
```
gmao-rreichle: Not sure we want/need date/time.  That's in the file stamp.  Do we have date/time in MAPL HISTORY nc4 output?  I have no idea.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'exp_id', trim(exp_id)) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'file_tag', trim(file_tag)) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'out_ObsFcstAna_format', trim(out_ObsFcstAna_format)) )
```
gmao-rreichle: "file_tag" is part of the file name and probably not needed.  I'm wondering how we could explain to a general user what "file_tag" means.  Currently, I don't see anything like a LONG_NAME for this attribute (if that even makes sense)
I don't think we need "out_ObsFcstAna_format" as an Attribute.  Hard to explain what it means, and if the user is looking at an nc4 file, they already know the format is nc4

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'update_type', update_type) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'dtstep_assim', dtstep_assim) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'grid_name', trim(grid_name)) )
```
gmao-rreichle: The value of this in the sample output seemed inconsistent with the nomenclature we use for the EASE tile space in the shared bcs archive.  It's likely that the source code doesn't fully conform to the nomenclature.  This requires more thought.  If in doubt, we may not want to keep this attribute. 
But this brings up another issue: HISTORY output in tile space is meant to be used with together with the "tile_coord" output file (for more complete coordinate info).  For M21C-Land, we might also need to provide the tilecoord output in nc4 format :(

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+
+    do i=1,N_obs_param
+       write(attr_name, '(A,I4.4,A)') 'species_', obs_param(i)%species, '_descr'
```
gmao-rreichle: Maybe format the integer to always have three digits (allowing up to 999 species)?  Then the strings line up better and are easier to process if anyone ever needs this

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1689,60 +1702,319 @@ subroutine output_ObsFcstAna(date_time, exp_id, &
...
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'N_obs_param', N_obs_param) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'year',    date_time%year) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'month',   date_time%month) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'day',     date_time%day) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'hour',    date_time%hour) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'minute',  date_time%min) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'second',  date_time%sec) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'dofyr',   date_time%dofyr) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'pentad',  date_time%pentad) )
+    call nc4_check( nf90_put_att(ncid, NF90_GLOBAL, 'exp_id', trim(exp_id)) )
```
gmao-rreichle: It's good to have the exp id, but should it be limited to 40 chars?  I don't know of the top of my head what GEOSldas supports, but it may be longer than 40 chars

gmao-rreichle (commented):
@amfox37 : Thanks again for putting together this PR.  I went through everything fairly quickly and made a bunch of comments for further discussion.  Unless you think something is obvious, no need to change anything until we've had a chance to discuss at our next tagup.

GEOSldas_App/lenkf_j_template.py
amfox37: Good point. Gone ahead and removed this block and removed filename extension from existing block.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: No, I don't think so we do. e.g. *.inst3_1d_lndfcstana_Nt.20150430.nc4 just has CreatedBy and Date (when written) in global attributes.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: Something like 
vname = strtrim(string(cellstr(ncread(fname,'species_varname')'))); 
does seem to work for me, but I'll investigate writing as a string.

amfox37: I did originally have them as global attrs but decided a  structured table data  that aligns with species_id/assim/scale/getinnov/errstd is better - but willing to be convinced otherwise.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: I don't think this is the case. The 40-char limits apply to obs_param text fields like varname/units/descr. exp_id is declared as character(len=ESMF_MAXSTR) in GEOS_LandAssimGridComp.F90 and nf90_put_att(..., 'exp_id', trim(exp_id)) writes the full trimmed string. I don't think its limited anywhere else.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: OK, let's switch to 3-digit zero-padding for readability, and the AI suggests using I0.3 so it still behaves safely if species IDs ever exceed 999.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: Switched NF90_STRING species text writes to MAPL pfio_nf90_put_var_string helper calls and I think this is now working OK.  A straight forward vname = ncread(fname,'species_varname'); now works.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: AI says:

Yes, NC4 has no logical type — the conversion is already handled explicitly: assim_int is allocated as an integer array, initialised to 0, then set to 1 where %assim is .true. via a where construct before the nf90_put_var call. Same pattern for species_assim_int.

amfox37: I removed the redundant N_obsf/N_obs_param global attributes and now rely on the nobs and n_species dimensions as the authoritative counts for consistency

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: I’ve renamed the fields derived from obs_param to an obsparam_ prefix so folsk can tell these are observation-parameter metadata.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: I removed the redundant N_obsf/N_obs_param global attributes and now rely on the nobs and n_species dimensions as the authoritative counts for consistency

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: Changed to assim_flag

GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
amfox37: Removed the duplication with a new subroutine

GEOSldas_App/LDASsa_DEFAULT_inputs_ensupd.nml
amfox37: Gone ahead an implemented out_ObsFcstAna = 0 etc. Note, this means they are not backward compatible, but I've tried adding what is hopefully an informative error message, although will be triggered by any namelist parse error, not just someone trying to input a logical here.

GEOSlandpert_GridComp/random_fields.F90
amfox37: This is a bit of a mess with this PR as I originally branched from the MKL fix branch, then removed those commits but it then doesn't merge correctly with Develop with those earlier commits. I've add them back in now so hopefully it reflects develop.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: Removed N_obs_param

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: Have remove the assim window time stamp information and added a Date (when written) and CreatedBy

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: Removed these

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: I think this is being handled by MAPL elsewhere.

GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
amfox37: OK, I've removed grid_name for now to avoid ambiguous metadata.

Guess we can have a follow up PR to additionally also ouput tilecoord.bin as a netcdf file

amfox37:
@gmao-rreichle I think I've had a go at addressing everything we discussed yesterday so probably time for you have a look through. Still need to tweak some names here and there maybe?

GEOSldas_App/LDASsa_DEFAULT_inputs_ensupd.nml
gmao-rreichle: @amfox37 : I cleaned up the config variables -- there were three of them that contained essentially the same info.  The intent was to make only 0-diff changes.

gmao-rreichle (commented):
@amfox37 : I cleaned up a few things.  My intent was to make only 0-diff changes.  When you get a chance, please verify that everything still works as intended. Also, please let me know where to find a sample ObsFcstAna output file in nc4 format.

amfox37:
@gmao-rreichle : Yes, this still seems to be producing .nc files which include the same content as the .bin files. You can find them here - `/discover/nobackup/projects/land_da/all_sensors/LS_DAv8_M36_as_test2/output/SMAP_EASEv2_M36_GLOBAL/ana/ens_avg/Y2020/M05`

gmao-rreichle:
>`/discover/nobackup/projects/land_da/all_sensors/LS_DAv8_M36_as_test2/output/SMAP_EASEv2_M36_GLOBAL/ana/ens_avg/Y2020/M05`

@amfox37 : This file has CYGNSS s.m. units as "cm3 cm-3", which is probably a leftover from an outdated "special" ens update nml file.  The "default" nml file has the correct (standard) "m3 m-3" units.  Please make sure to change this on your end so the non-standard units don't find their way into M21C-Land.

GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -358,11 +358,11 @@ subroutine read_ens_upd_inputs(               &
 
        ! expect "fcstvarname" and "fcstunits" to be 'NULL'; they are not meant to be filled by the user in the config nml file
        
-       if ( .not. ( (trim(obs_param_nml(i)%fcstvarname) /= 'NULL') .or. (trim(obs_param_nml(i)%fcstvarname) /= 'null') ) )     &
+       if ( .not. ( (trim(obs_param_nml(i)%fcstvarname) /= 'NULL') .and. (trim(obs_param_nml(i)%fcstvarname) /= 'null') ) )     &
```
gmao-rreichle: @amfox37 :  Sorry for the mess here.  I think we also need to change both instances of `/=` to `==`.  What we really want (in pseudocode) is:
`if uppercase(fcstvarname)/='NULL', abort`
I'm using the complicated logic b/c I'm not aware of a Fortran function for "uppercase", but I do want to allow lower-case "null".  (I'm being lazy on mixed-case such as "Null").  Maybe there's something in MAPL?

amfox37: @gmao-rreichle: Actually, think we just need

`if (trim(fcstvarname) /= 'NULL' .and. trim(fcstvarname) /= 'null') then
   call abort
endif`

I'll check this

amfox37: Yes, this is now giving the desired behaviour I believe. I did test putting in non-NULL values. As before, the example ouputs are here: `/discover/nobackup/projects/land_da/all_sensors/LS_DAv8_M36_as_test2/output/SMAP_EASEv2_M36_GLOBAL/ana/ens_avg/Y2020/M05`

amfox37:
These last few changes seem fine. Everything built and ran and the checks comparing with values in the .bin files passed.

gmao-rreichle:
Hi @mathomp4 and @weiyuan-jiang :  I have a weird Fortran question.  In a recent commit 4aecf203067758cf27067d69b8d2d991a429ef8d , I implemented a change of no-data-value using a vectorized "where" together with the elemental function "LDAS_is_nodata()": 
https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/6d533a1cb3d35f2e6c1266056a5aaf8bc1728a30/LDAS_Shared/LDAS_ensdrv_Globals.F90#L134-L143
When we tested this, the -9999s in the Observations structure were sometimes correctly converted into 1e15 and sometimes not. 
Andy found that if we replace the vectorized where statement with a do loop, as in 41ab7d83139aa7a71070cbd7a35cc10ffbfdeb73, the no-data-value conversion works as expected.  
The AI is telling us we did nothing wrong, and we're stumped.  But it didn't work, so something must be wrong.  Any ideas what the problem might be? 
We can just leave the do loop as it is now since it's working, but I'd still love to understand what went wrong with the vectorized statements.
Thanks! 

cc: @amfox37

weiyuan-jiang:
> Hi @mathomp4 and @weiyuan-jiang : I have a weird Fortran question. In a recent commit [4aecf20](https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/4aecf203067758cf27067d69b8d2d991a429ef8d) , I implemented a change of no-data-value using a vectorized "where" together with the elemental function "LDAS_is_nodata()":
> 
> https://github.com/GEOS-ESM/GEOSldas_GridComp/blob/6d533a1cb3d35f2e6c1266056a5aaf8bc1728a30/LDAS_Shared/LDAS_ensdrv_Globals.F90#L134-L143
> 
> 
> When we tested this, the -9999s in the Observations structure were sometimes correctly converted into 1e15 and sometimes not.
> Andy found that if we replace the vectorized where statement with a do loop, as in [41ab7d8](https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/41ab7d83139aa7a71070cbd7a35cc10ffbfdeb73), the no-data-value conversion works as expected.
> The AI is telling us we did nothing wrong, and we're stumped. But it didn't work, so something must be wrong. Any ideas what the problem might be?
> We can just leave the do loop as it is now since it's working, but I'd still love to understand what went wrong with the vectorized statements.
> Thanks!
> cc: @amfox37

I think the code is correct. All I can think of is the compiler's bug. We can see if this bug is reproducible.

tclune:
I agree - pretty hard to be "slightly" wrong when using an elemental function in this manner.   But it is also a rather surprising compiler bug.   And as we want to do more things like this, it is probably worth submitting a reproducer.  (Unless fixed in more recent compiler versions.)

weiyuan-jiang:
> I agree - pretty hard to be "slightly" wrong when using an elemental function in this manner. But it is also a rather surprising compiler bug. And as we want to do more things like this, it is probably worth submitting a reproducer. (Unless fixed in more recent compiler versions.)

Let me try if I can write a simple example to reproduce the error. It may be hard since Rolf said sometimes it produced correct answers

gmao-rreichle:
@tclune @weiyuan-jiang : Thanks for taking a look, much appreciated.  Let me clarify the "sometimes" a bit more.  The data we're writing are fields of a structure, "Observations".  If I understand @amfox37's report correctly, the field "ana" did not convert the no-data-value properly (it left the -9999s in place), whereas the -9999s were converted to 1e15 as desired for the field "anavar".   This result was obtained using something like the LDAS_GLOBAL/assim "nightly" regression test.  It might be tricky to reproduce the bug.  Perhaps start with a simple script that initializes the "Observations" structure with random numbers, put -9999s in random places, and then do the no-data-value conversion in the vectorize and loop implementations of https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/41ab7d83139aa7a71070cbd7a35cc10ffbfdeb73

weiyuan-jiang:
I let Claud write a program and test it under gfortran, ifort, ifx and nag. All pass.

program main
  implicit none
  real, parameter :: nodata_generic         = -9999.
  real, parameter :: nodata_tolfrac_generic = 1.e-4
  real, parameter :: MAPL_UNDEF = 1.0E15

  real :: nodata_tol_generic     = abs(nodata_generic*nodata_tolfrac_generic)
  real :: MAPL_UNDEF_tol_generic = abs(MAPL_UNDEF    *nodata_tolfrac_generic)

  integer :: N_obsf, i
  real, allocatable :: tmpvecreal(:)
  real :: r
  integer :: n_replaced, n_undef_before, n_nodata_before

  N_obsf = 1000
  allocate(tmpvecreal(N_obsf))

  ! Fill tmpvecreal with random mix of -9999., MAPL_UNDEF, and normal values
  call random_seed()
  do i = 1, N_obsf
     call random_number(r)
     if (r < 0.33) then
        tmpvecreal(i) = nodata_generic   ! -9999.
     else if (r < 0.66) then
        tmpvecreal(i) = MAPL_UNDEF       ! 1.0E15
     else
        tmpvecreal(i) = r * 100.0        ! some normal value
     end if
  end do
 ! Count before replacement
  n_nodata_before = count(abs(tmpvecreal - nodata_generic) < nodata_tol_generic)
  n_undef_before  = count(abs(tmpvecreal - MAPL_UNDEF)     < MAPL_UNDEF_tol_generic)

  print *, "--- Before replacement ---"
  print *, "  Elements equal to -9999.    :", n_nodata_before
  print *, "  Elements equal to MAPL_UNDEF:", n_undef_before
  print *, "  Other (normal) values       :", N_obsf - n_nodata_before - n_undef_before

  ! Replace -9999. (and any existing MAPL_UNDEF) with MAPL_UNDEF
  where (LDAS_is_nodata(tmpvecreal)) tmpvecreal = MAPL_UNDEF

  ! Count after replacement
  n_replaced = count(abs(tmpvecreal - MAPL_UNDEF) < MAPL_UNDEF_tol_generic)

  print *, ""
  print *, "--- After replacement ---"
  print *, "  Elements equal to MAPL_UNDEF  :", n_replaced
  print *, "  Elements still equal to -9999.:", count(abs(tmpvecreal - nodata_generic) < nodata_tol_generic)

  if (n_replaced == n_nodata_before + n_undef_before) then
     print *, ""
     print *, "SUCCESS: all -9999. and MAPL_UNDEF entries replaced with MAPL_UNDEF."
  else
     print *, ""
     print *, "FAILURE: replacement count does not match expected."
  end if

  deallocate(tmpvecreal)
contains

   elemental function LDAS_is_nodata(data) result(no_data)

      real,   intent(in) :: data
      logical            :: no_data

      no_data =                                                         &
           ( abs(data-nodata_generic) < nodata_tol_generic    ) .or.    &
           ( abs(data-MAPL_UNDEF)     < MAPL_UNDEF_tol_generic)

    end function LDAS_is_nodata


end program

weiyuan-jiang:
I also ask Claude to put some Nans into the tmpvecreal. The Nan will not be replaced.

tclune:
> @tclune @weiyuan-jiang : Thanks for taking a look, much appreciated. Let me clarify the "sometimes" a bit more. The data we're writing are fields of a structure, "Observations". If I understand @amfox37's report correctly, the field "ana" did not convert the no-data-value properly (it left the -9999s in place), whereas the -9999s were converted to 1e15 as desired for the field "anavar". This result was obtained using something like the LDAS_GLOBAL/assim "nightly" regression test. It might be tricky to reproduce the bug. Perhaps start with a simple script that initializes the "Observations" structure with random numbers, put -9999s in random places, and then do the no-data-value conversion in the vectorize and loop implementations of [41ab7d8](https://github.com/GEOS-ESM/GEOSldas_GridComp/commit/41ab7d83139aa7a71070cbd7a35cc10ffbfdeb73)

Yes - I only meant to create a synthetic reproducer.  Reproducing in the full run would give little traction.   Of course, sometimes synthetic reproducers don't (reproduce).

gmao-rreichle:
Thanks, @weiyuan-jiang, for giving this a shot.  Another possible try could be to have a "reproducer" that reads the "Observations" data from the existing binary file whose nc4 twin didn't have the right no-data-values after conversion.  (Andy's test runs produce both binary and nc4 output.). That might reveal any issues with  the actual numbers.

More likely, however, Claude's "reproducer" doesn't reproduce because it doesn't use the "Observations" structure and my implementation of the no-data-value conversion of multiple fields.  Note that my vectorized code repeatedly assigns fields from the structure to "tmpvecreal".  Perhaps this repeated use of "tmpvecreal" gets out of sync when the executable is built with some optimization?

weiyuan-jiang:
> Thanks, @weiyuan-jiang, for giving this a shot. Another possible try could be to have a "reproducer" that reads the "Observations" data from the existing binary file whose nc4 twin didn't have the right no-data-values after conversion. (Andy's test runs produce both binary and nc4 output.). That might reveal any issues with the actual numbers.
> 
> More likely, however, Claude's "reproducer" doesn't reproduce because it doesn't use the "Observations" structure and my implementation of the no-data-value conversion of multiple fields. Note that my vectorized code repeatedly assigns fields from the structure to "tmpvecreal". Perhaps this repeated use of "tmpvecreal" gets out of sync when the executable is built with some optimization?

I would like a full model run to figure out why. We can focus on the incorrect replacements , print out the vectorized results and the do-loop results and see what are the differences.

gmao-rreichle:
> I would like a full model run to figure out why. We can focus on the incorrect replacements , print out the vectorized results and the do-loop results and see what are the differences.

@amfox37 : Please point @weiyuan-jiang to your run and let him know which of the bin/nc files show the issue?  Thanks

mathomp4:
@weiyuan-jiang Can you try ifort again using the same compiler flags we use in GEOS. Maybe one of the weird ifort flags causes an issue?

Our ifort flags would be:
```
-O3 -g -march=core-avx2 -fma -align array32byte -traceback -assume realloc_lhs -qopt-report0 -align all -fno-alias -fpe3 -fp-model fast -fp-model source -fp-model consistent -ftz -assume noold_maxminloc -diag-disable 10121 -diag-disable=10448 -align dcommons -fPIC -qopenmp
```

weiyuan-jiang:
> @weiyuan-jiang Can you try ifort again using the same compiler flags we use in GEOS. Maybe one of the weird ifort flags causes an issue?
> 
> Our ifort flags would be:
> 
> ```
> -O3 -g -march=core-avx2 -fma -align array32byte -traceback -assume realloc_lhs -qopt-report0 -align all -fno-alias -fpe3 -fp-model fast -fp-model source -fp-model consistent -ftz -assume noold_maxminloc -diag-disable 10121 -diag-disable=10448 -align dcommons -fPIC -qopenmp
> ```

I tried those flags and it still worked

gmao-rreichle (approved):

amfox37:
> > I would like a full model run to figure out why. We can focus on the incorrect replacements , print out the vectorized results and the do-loop results and see what are the differences.
> 
> @amfox37 : Please point @weiyuan-jiang to your run and let him know which of the bin/nc files show the issue? Thanks

Hi @weiyuan-jiang, if you want to take a look at outputs look at **test2**: `/gpfsm/dnb06/projects/p284/all_sensors/LS_DAv8_M36_as_test2/output/SMAP_EASEv2_M36_GLOBAL/ana/ens_avg/Y2020/M05/`

6 out of the 16 .nc4 ObsFcstAna files have the issue and differ from the .bin files. Specifically:
20200501_1200z, 20200502_0000z, 20200502_0600z, 20200502_0900z, 20200502_2100z & 20200503_0000z

I built and re-ran several times with various debug statements and always got the same error in the same files. 

**test3** is with the do-loops (and run out for a few more days) `gpfsm/dnb06/projects/p284/all_sensors/LS_DAv8_M36_as_test3`

weiyuan-jiang:
To reproduce the error, I tried to write out the data. But with the write statement, -9999s disappear and everything looks fine. Maybe we should try nag compiler on GEOSldas.

       tmpvecreal = Observations_f(1:N_obsf)%ana
       open(unit=10, file='tmpvecreal.txt', status='replace', action='write')
       do i=1,N_obsf
         write(10, *) tmpvecreal(i)
       end do
       close(10)
       where (LDAS_is_nodata(tmpvecreal)) tmpvecreal=MAPL_UNDEF; call nc4_check( nf90_put_var(ncid, ana_varid,     tmpvecreal))
       do i=1,N_obsf
          if (tmpvecreal(i)<-9998) then
            print*, tmpvecreal(i), i, n_obsf, ldas_is_nodata(tmpvecreal(i))
            stop "damned, still have -9999."
          endif
       end do
       
@amfox37 @gmao-rreichle

gmao-rreichle:
> To reproduce the error, I tried to write out the data. But with the write statement, -9999s disappear and everything looks fine. Maybe we should try nag compiler on GEOSldas.
> 
> ```
>    tmpvecreal = Observations_f(1:N_obsf)%ana
>    open(unit=10, file='tmpvecreal.txt', status='replace', action='write')
>    do i=1,N_obsf
>      write(10, *) tmpvecreal(i)
>    end do
>    close(10)
>    where (LDAS_is_nodata(tmpvecreal)) tmpvecreal=MAPL_UNDEF; call nc4_check( nf90_put_var(ncid, ana_varid,     tmpvecreal))
>    do i=1,N_obsf
>       if (tmpvecreal(i)<-9998) then
>         print*, tmpvecreal(i), i, n_obsf, ldas_is_nodata(tmpvecreal(i))
>         stop "damned, still have -9999."
>       endif
>    end do
> ```
> 
> @amfox37 @gmao-rreichle

Thanks, @weiyuan-jiang, for giving this another try.  The only potentially relevant difference that I can still see between the code snippet above and the original (vectorized) implementation is that above you're only converting/writing one field ("ana").  In the original vectorized implementation, we repeated the convert/write sequence for the 6 science data fields ("obs", "obsvar", "fcst", "fcstvar", "ana", "anavar"), with no other statements in-between subsequent convert/write statements, as in 

```
 ! for data fields, replace LDAS no-data-value with MAPL_UNDEF for consistency with MAPL HISTORY output
       tmpvecreal = Observations_f(1:N_obsf)%obs    ; where (LDAS_is_nodata(tmpvecreal)) tmpvecreal=MAPL_UNDEF; call nc4_check( nf90_put_var(ncid, obs_varid,     tmpvecreal))
       tmpvecreal = Observations_f(1:N_obsf)%obsvar ; where (LDAS_is_nodata(tmpvecreal)) tmpvecreal=MAPL_UNDEF; call nc4_check( nf90_put_var(ncid, obsvar_varid,  tmpvecreal))
       tmpvecreal = Observations_f(1:N_obsf)%fcst   ; where (LDAS_is_nodata(tmpvecreal)) tmpvecreal=MAPL_UNDEF; call nc4_check( nf90_put_var(ncid, fcst_varid,    tmpvecreal))
       tmpvecreal = Observations_f(1:N_obsf)%fcstvar; where (LDAS_is_nodata(tmpvecreal)) tmpvecreal=MAPL_UNDEF; call nc4_check( nf90_put_var(ncid, fcstvar_varid, tmpvecreal))
       tmpvecreal = Observations_f(1:N_obsf)%ana    ; where (LDAS_is_nodata(tmpvecreal)) tmpvecreal=MAPL_UNDEF; call nc4_check( nf90_put_var(ncid, ana_varid,     tmpvecreal))
       tmpvecreal = Observations_f(1:N_obsf)%anavar ; where (LDAS_is_nodata(tmpvecreal)) tmpvecreal=MAPL_UNDEF; call nc4_check( nf90_put_var(ncid, anavar_varid,  tmpvecreal))
```

I don't see why the sequence of convert/write statements should behave any differently, but it does when it's part of the full GEOSldas executable.  Maybe try the full sequence in the "reproducer"?  If that doesn't work, maybe we should drop this, at least for now.

PR #166, opened 2026-03-20
clean up MERRA-2 restart option 

weiyuan-jiang:
This PR simplified the check on RESTART='M'. 

Related PRs:
https://github.com/GEOS-ESM/GEOS_Util/pull/188

### DIFF
--- a/GEOSldas_App/ldas.py
+++ b/GEOSldas_App/ldas.py
@@ -819,20 +819,24 @@ class ldas:
            while bcs_path[-1] == '/' : bcs_path = bcs_path[0:-1]
            bc_base    = os.path.dirname(bcs_path)
            bc_version = os.path.basename(bcs_path)
-
-           remap_tpl = os.path.dirname(os.path.realpath(__file__)) + '/remap_params.tpl'
-           config = yaml_to_config(remap_tpl)
+           answers={}
+           if RESTART_str == 'M' : 
+              answers['input:shared:MERRA-2']   = True
+              answers['input:shared:yyyymmddhh']= YYYYMMDDHH
+              answers['output:shared:out_dir']  = mk_outdir
+              init_merra2(answers)
+           config = get_config_from_answers(answers, config_tpl = True)
 
            config['slurm_pbs']['account'] = self.RmInputs['account']
            config['slurm_pbs']['qos'] = 'debug'
 
-           config['input']['surface']['catch_tilefile']  = self.in_tilefile
-           config['input']['shared']['expid']            = self.ExeInputs['RESTART_ID']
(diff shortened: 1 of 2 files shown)


GEOSldas_App/ldas.py
```diff
@@ -845,33 +849,6 @@ def createLnRstBc(self) :
...
-             elif (yyyymm < 201101):
-                merra2_expid = "d5124_m2_jan00"
-             elif (yyyymm < 202106):
-                merra2_expid = "d5124_m2_jan10"
-             # There was a rewind in MERRA2 from Jun 2021 to Sept 2021
-             elif (yyyymm < 202110):
-                merra2_expid = "d5124_m2_jun21"
-             config['input']['shared']['expid']      = merra2_expid
-             config['input']['shared']['rst_dir']    = mk_outdir+ '/merra2_tmp_'+ YYYYMMDDHH
-             config['input']['surface']['wemin']     = 26
```
gmao-rreichle, line 866: @weiyuan-jiang : In GEOSldas, for RESTART='M' we used to explicitly define WEMIN=26.  The function init_merra2() in remap_restarts does not do this.  Are we sure that the bcs-specific default of 26 for bcs_version='GM4' kicks in when we remap from MERRA-2?  I guess it's a question of the order of operations.  As long as 'input:shared:bc_version' is defined before wemin_default() kicks in, we should be ok.  But at least for me that's not easy to assert by just looking at the code.

weiyuan-jiang: You are right. Added [here](https://github.com/GEOS-ESM/GEOS_Util/pull/189) > it may be duplicated at some point but it is safer.

biljanaorescanin:
**Testing summary:**

PR was tested in combination with [GEOS_Util 188](https://github.com/GEOS-ESM/GEOS_Util/pull/188)
` All regression tests passed. `

 **For functionality:** 
Testing M2 RESTART option from **Discover36 worked**.
If we try same setup from **Discover32** user will see:

 ` File "/gpfsm/dnb34/borescan/brisi/pr188_util/GEOSldas/install/bin/remap_utils.py", line 155, in init_merra2`
  `  assert os.path.exists(MERRA2_RST_BASE), "Must be on discover30 or discover36 to access MERRA-2 ``restarts at " + MERRA2_RST_BASE`
  `         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^`
`AssertionError: Must be on discover30 or discover36 to access MERRA-2 restarts at ``/archive/users/gmao_ops/MERRA2/gmao_ops/GEOSadas-5_12_4/`

gmao-rreichle (dismissed):

gmao-rreichle (approved):

PR #174, opened 2026-05-18
add stop during setup when running routing but domain is not global

gmao-rreichle:


GEOSldas_App/ldas.py
```diff
@@ -763,6 +764,10 @@ def createLnRstBc(self) :
            self.isZoomIn= True
         #os.remove(self.domain_def.name)
 
+        # if running routing, make sure domain is global
+        if self.run_route>0 and self.isZoomIn=True:
```
weiyuan-jiang:
```suggestion
        if self.run_route>0 and self.isZoomIn:
```

biljanaorescanin (approved):

PR #177, opened 2026-06-23
HOTFIX, updated LADAS default setup setting and resource files

gmao-qliu:
CLOSED 24 Jun 2026: Replaced by #178 

------

This pull request introduces the following updates and fixes:

Replaced the old scaling parameters (which were based on the corrected "M21CLFO" precipitation) with new parameters derived from offline LDAS simulations using "M21CINT" precipitation forcing.

Updated the new default configuration to ensure the mwRTM "clay" parameter is consistent with NLv3.

Resolved the ref_time issue for HISTORY.rc.

Modified the default "ntasks" from 120 to 136 to ensure more than 1 node will be used. 

In job script, remove "lmc", "obsparam", and "nml" files generated by GEOSldas.x when coupled to ADAS to avoid crash after rewind.


weiyuan-jiang (commented):
Please update CHANGELOG.md

gmao-rreichle:
@gmao-qliu : Thanks for putting this together.  I added a couple of commits that improve comments/documentation.  
Also, the name of the target branch should probably be "release/v3.1.1".   Otherwise, after the merge, the branch  labeled "v3.1.0" would contain what we want to be v3.1.1, which is very confusing.   But let me check with the SI team first about how exactly I should create a hot fix release, and what the naming conventions are for the branches.

PR #181, opened 2026-07-08
Add lake tile type

biljanaorescanin:
This PR is part of 3 PRs that help us run just lake tiles in GEOSldas offline mode. 

1. Add Lake offline mode. 
2. Add lake HISTORY ( 1D and 2D collections) and Restart.  (NOTE: maybe 1D collection should be removed if we don't enable EASE grid runs.) 

**Functionality testing:** 
Tested for run `tile_type 100 20` (zero diff to current global and globalcs tests).
Tested for `tile_type 100 19 20` we can only do `globalcs` for now.  Test was done for c360. 

Plumbing now works and the remaining risk is stability and science sanity, working on longer run for validation. 
Preliminary RST magnitude and spatial comparisons for offline vs coupled 2day run are at same locations with very similar range. 

https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1470
https://github.com/GEOS-ESM/GEOSldas/pull/862


GEOS_LdasGridComp.F90
gmao-rreichle: @weiyuan-jiang : Do you remember why we need this particular run_method ("DistributeForcingToLand")?  It's connecting aerosol forcings plus PS and DZ between METFORCE and LAND, but nothing else.

GEOSensavg_GridComp/GEOS_LandiceAvgGridComp.F90
gmao-rreichle: @lcandre2 : This now matches M21C specs but doesn't have the "glacier" qualifier in the LONG_NAME.  I wonder if we should add such a qualifier in the LONG_NAME.  In some ways, it's implicit in the fact that variables pulled from here into the "glc" collection are glacier-only quantities.  But users get confused easily, and extra clarity might be helpful.  Adding "glacier" (or similar) to the LONG_NAME would be a departure from M21C file specs.

gmao-rreichle (commented):
@biljanaorescanin, @weiyuan-jiang, @lcandre2, @agstub  : 

I did my best to fix up some missing functionality in the GEOSldas GC portion of the Lake PR.  Specifically, we need Lake to run with the Land **ensemble**.  

We already decided to run only one member for Landice.  Lake doesn't have the same computational constraints as Landice+ISSM, but I think we should also limit Lake to one member.  After all, if we added an ensemble of Lake GCs, it would not be used at all (no assimilation or uncertainty output).  So I deleted the hard stop triggered by "with_lake==.true. .and. N_ens>1" and tried to make Lake behave like Landice as far as ens members are concerned.  @weiyuan-jiang, please review the code changes.

It may not have been technically needed, but I added a LakeAvg GC.  @lcandre2, I think the [Land,Lake,Landice]Avg GridComp give us an opportunity to change LONG_NAME attributes from those in the GCM GC, and I took advantage of that for Lake. 

I forget if we figured out how the "native" ISSM outputs (i.e., the "issm" collection) will behave when we run in ensemble mode.  It's probably best to just try it out.  I also didn't try to use the "HAVE_ISSM" preprocessor directive in the LDAS GC. @agstub, I may need some help if this doesn't work as is.

@biljanaorescanin : Please try to set up a (c90?) test run for a few days with NUM_ENSEMBLE=4 (and no data assimilation for now).  Write out the "lnd", "lake", "glc", and "issm" collections defined in GEOSldas_HIST.rc.  The "source" GridComp specs might need to change to LAKEAVG, LANDICEAVG, etc to make sure we're pulling from the "AVG" Grid Comps and not the LAKE, LANDICE, etc GCs.  (This will need to be cleaned up in the script that preprocesses GEOSldas_HIST.rc)  Also, for now you can just make four identical copies of a single restart and use them to restart the ensemble (or it may still be possible to restart an ens simulation from a single member restart that is then used for all members).  I'm sure I'm forgetting lots of details, so please be in touch with questions.  Thanks! 

Also, see my comments below.

GEOS_LdasGridComp.F90
weiyuan-jiang: We need this call because metforce may be on other types of tiles ( not just land). This call is for variables that are not perturbed. So I guess those variables do not need to be perturbed.

weiyuan-jiang:
quote: "So I deleted the hard stop triggered by "with_lake==.true. .and. N_ens>1" and tried to make Lake behave like Landice as far as ens members are concerned."

That looks fine with me.

biljanaorescanin:
All requested regression tests passed.

gmao-rreichle:
@mathomp4 : Now that #194 is in, I restored ifx in the workflow (https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1510).  I hope I got this right.  When you get a chance, please look over the CI changes one more time to see if they're how would want them to look before we merge the PR (which we're ready to do once Scott has merged GCM GC PR 1510).  Thanks

GEOSensavg_GridComp/GEOS_LandiceAvgGridComp.F90
gmao-rreichle: Resolving this comment so we can merge.   If needed, we can make changes in a new PR.

gmao-rreichle (approved):

gmao-rreichle (approved):

gmao-rreichle (approved):

PR #186, opened 2026-07-21
Add H SAF ASCAT H121/H139 soil moisture reader

amfox37:
This PR adds support for reading H SAF ASCAT surface soil moisture from the H121 CDR v8 and H139 ICDR NetCDF products.

Summary:
- Adds `read_obs_sm_ASCAT_HSAF()` for H SAF ASCAT NetCDF swath files on the 12.5 km Fibonacci grid.
- Adds MetOp-A/B/C H SAF ASCAT observation species:
  - `ASCAT_HSAF_META_SM`
  - `ASCAT_HSAF_METB_SM`
  - `ASCAT_HSAF_METC_SM`
- Increases `N_obs_species_nml` from 55 to 58.
- Adds default ensupd namelist entries for H121/H139 data paths and flists.
- Wires the new H SAF ASCAT species into `read_obs()`.
- Applies H SAF-specific QC using surface/processing flags, wetland fraction, topographic complexity, subsurface scattering probability, soil moisture sensitivity, and noisy backscatter flags.
- Leaves snow/frozen screening to the existing model-based satellite soil moisture QC.
- Updates `CHANGELOG.md`.

Notes:
- MetOp-A is configured for the H121 CDR period.
- MetOp-B/C are configured for the H121 CDR plus H139 ICDR continuity path.

Testing:
- Passed GEOSldas regression tests on Discover.
- Used extensively in multi-year experiments on Discover with real H SAF ASCAT H121/H139 input.
- Built locally on macOS with `cmake --build build-develop-20260520 --target install -j 4`; the latest pushed commit is changelog/comment-only.


gmao-rreichle:
@amfox37 : Thanks for putting this together, looks great.  I added a few commits, mostly comments and white-space changes for cleanup.  

There is, however, one change that needs discussion: I think the FOV value should be 12.5 km: 3f267f33dea756e36cd31823e1bfb65c6adf0ac0.  The FOV value is a **radius**.  SMAP Tb has ~40 km resolution, and the FOV value is 20 km.  The "12.5km" ASCAT product a nominal resolution of 25 km per Hahn et al, so FOV should be 12.5 km.  

Not sure what the resolution of the old EUMETSAT product is.  Probably the FOV value of 20 km isn't quite right, but we can leave that alone given that we're going to use the HSAF product going forward.  

Does this make sense to you?

GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -2148,6 +2148,525 @@ end subroutine read_obs_sm_ASCAT_EUMET
...
+    real,    intent(out), dimension(N_catd) :: ASCAT_lon, ASCAT_lat
+    real*8,  intent(out), dimension(N_catd) :: ASCAT_time             ! J2000 seconds
+
+    ! ---------------
+
+    ! H121/H139 files are 1-hour granules.  Extra look-back catches files
+    ! whose sensing start precedes the window but whose obs extend into it.
+
+    integer,      parameter :: dt_ASCAT_obsfile = 3600                ! seconds (1-hour files)
+    integer,      parameter :: N_fnames_max     = 30                  ! max obs files per daily flist
```
gmao-rreichle: I'm a bit confused why the number here is 30.  With ~100 minutes in each file, there should be  no more than 15 files per day (incl. accounting for the case when the first or last file doesn't start or end on a day boundary). 
Why isn't N_fnames_max=15? 
It is good practice to use the smallest possible max value.  In SMAP ops, we occasionally have issues with stray or duplicate files.  A tight N_max can help weed out such cases.

amfox37: Hmm. They are hourly so actually I think this should be 24. Let me double check.

gmao-rreichle: Sorry, no idea where I thought I saw 100 min.  The files do seem to cover exact hours, starting at 0z.  So it looks like 24 is the correct number, then.

amfox37: The previous ASCAT obs were grouped in ~100 minute segments.

gmao-rreichle: Changed to 24 in commit 9d05ba52815b3735443bc9583a219e8af779540f

amfox37:
Added the peatland QC in get_obs_pred() for sfsd and smfc observation types.

